PR #9, opened 2022-06-29
List the latest 100 cases for the users' force

C-gyorfi:


src/useCases/listCases.ts
```diff
@@ -0,0 +1,39 @@
+import Database from "types/Database"
+import ErrorListItem from "types/ErrorListItem"
+import PromiseResult from "types/PromiseResult"
+
+const listCases = (connection: Database, forces: string[], limit: number): PromiseResult<ErrorListItem[]> => {
```
bjpirt, line 5: We should be using variable substitution for these queries, not building strings, otherwise we end up being vulnerable to SQL injection attacks.

emadkaramad (approved):

PR #24, opened 2022-07-19
Bicaws 2154 filter by defendantname

joe-fol:
Add logic to query with a case-insensitive match.
Update listCourtCases to allow search by defendant name. 
PR for front end to follow

### DIFF
--- a/src/types/QueryParams.ts
+++ b/src/types/QueryParams.ts
@@ -4,5 +4,6 @@ export type QueryParams = {
   orderBy?: string
   order?: QueryOrder
   limit: number
+  defendantName?: string
   forces: string[]
 }
--- a/src/useCases/listCourtCases.ts
+++ b/src/useCases/listCourtCases.ts
@@ -1,9 +1,9 @@
-import CourtCase from "../entities/CourtCase"
 import { DataSource } from "typeorm"
-import PromiseResult from "../types/PromiseResult"
-import getColumnName from "../lib/getColumnName"
-import courtCasesByVisibleForcesQuery from "./queries/courtCasesByVisibleForcesQuery"
 import { QueryParams } from "types/QueryParams"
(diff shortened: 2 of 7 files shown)


src/types/QueryParams.ts
```diff
@@ -4,5 +4,6 @@ export type QueryParams = {
   orderBy?: string
```
emadkaramad, line 4: I would assume this file is only for the cases list page. Should we prefix the filename with the page name to be more specific?

emadkaramad (approved):

src/useCases/listCourtCases.ts
```diff
@@ -15,6 +15,12 @@ const listCourtCases = async (connection: DataSource, queryParams: QueryParams):
     )
     .limit(queryParams.limit)
 
+  if (queryParams.defendantName) {
+    query.where("lower(courtCase.defendantName) like '%' || :name || '%'", {
```
emadkaramad, line 19: For case-insensitive search in Postgres, it seems we can use [ILIKE](https://www.postgresql.org/docs/9.0/functions-matching.html) keyword. So we don't need to use `lower()` function.

C-gyorfi: Refactored in [b22d455](https://github.com/ministryofjustice/bichard7-next-ui/pull/24/commits/b22d455b17a38c3e89b99d3442fa71a1e35c2587)

src/types/QueryParams.ts
C-gyorfi: Refactored in [b22d455](https://github.com/ministryofjustice/bichard7-next-ui/pull/24/commits/b22d455b17a38c3e89b99d3442fa71a1e35c2587)

C-gyorfi (approved):

PR #26, opened 2022-07-19
Add filter for cases having triggers and exceptions

monotypical:
### In this PR:

- Adds an optional filter parameter to `listCourtCases` for retrieving only cases having exceptions or triggers. This is implemented using the `trigger_count` and `error_count` columns from the DB. This is tested with a jest integration test
- Adds a select dropdown to the case listing page which sets a URL parameter for the case filter, which is then passed to `listCourtCases`. This is tested with cypress, for which I also added an `insertException` task
- Tweaks the case listing cypress test to ensure trigger & exception codes are appearing in the right place on the page

### Notes

The dropdown UI isn't great, I spent a while trying to get the button adjacent to the dropdown but I couldn't figure it out and this is only a placeholder so I left it as-is

### Screenshots

no filter applied:
image

filtering cases with triggers:
image

filtering cases with exceptions:
image

### DIFF
--- a/src/components/CourtCaseFilter.tsx
+++ b/src/components/CourtCaseFilter.tsx
@@ -15,7 +15,8 @@ const CourtCaseFilter = (props: { initialSelection: Filter }) => {
           label={"Filter cases"}
           input={{
             value: currentSelection,
-            onChange: (e) => setCurrentSelection(queryParamToFilterState(e.target.value))
+            onChange: (e) => setCurrentSelection(queryParamToFilterState(e.target.value)),
+            id: "case-filter-select"
           }}
         >
           <option value={""} selected={!currentSelection}>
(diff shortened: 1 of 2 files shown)


src/components/CourtCaseFilter.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [currentSelection, setCurrentSelection] = useState(props.initialSelection)
+
+  return (
+    <GridRow>
+      <GridCol>
+        <Select
+          label={"Filter cases"}
+          input={{
+            value: currentSelection,
+            onChange: (e) => setCurrentSelection(queryParamToFilterState(e.target.value)),
```
C-gyorfi, line 18: I wonder if there is a neat way of using [form submission without JS](https://nextjs.org/docs/guides/building-forms#form-submission-without-javascript) so we don't need to use JS for filtering

C-gyorfi (approved):

PR #29, opened 2022-07-25
Add support for IE11

monotypical:
This PR:
* sets babel options to transpile the built client-side JS to be IE11-compatible using the `preset-env` plugin and a browsers list
* transpiles client-side modules using syntax too new for IE11
* adds polyfills for JS features we use which are missing from IE11

I've tested this on a local windows 7 VM for now, but plan to write automated tests using BrowserStack to run as part of the CI shortly

image

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,2 @@
+defaults
+ie 7-11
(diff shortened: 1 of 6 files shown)


jaskaransarkaria (approved):
🚀

PR #30, opened 2022-07-26
Paginate records

jaskaransarkaria:
* paginate cases when there are more than the max page items (add pagination component)
* add e2e test
* add integration test
* Add pagination story book
* Use query params for pagination


.storybook/main.js
```diff
@@ -1,3 +1,5 @@
+const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
```
jaskaransarkaria, line 1: this is needed to handle component path aliases for storybook

setupTests.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
jaskaransarkaria, line 1: eslint complains that this should in dev dependencies

cypress/support/e2e.ts
```diff
@@ -15,6 +15,8 @@
 
 // Import commands.js using ES2015 syntax:
 import "./commands"
+// eslint-disable-next-line import/no-extraneous-dependencies
```
jaskaransarkaria, line 18: eslint complains that this should in dev dependencies

.storybook/main.js
```diff
@@ -1,10 +1,13 @@
...
+
 module.exports = {
   stories: ["../src/**/*.stories.mdx", "../src/**/*.stories.@(js|jsx|ts|tsx)"],
   addons: [
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-interactions",
-    "storybook-addon-designs"
+    "storybook-addon-designs",
+    "storybook-addon-next-router"
```
tomvaughan77, line 10: Nice one 💯

tomvaughan77 (approved):

src/pages/index.tsx
```diff
@@ -14,79 +14,91 @@ import Head from "next/head"
...
+      pageNum: validatedPageNum,
+      orderBy: validatedOrderBy,
+      order: validatedOrder
     })
 
     if (isError(courtCases)) {
       throw courtCases
     }
 
-    const oppositeOrder: QueryOrder = order === "asc" ? "desc" : "asc"
```
C-gyorfi, line 56: Moving this before line 45, where we call `listCourtCases` causes a small bug, while the URL param shows asc/desc the return result is the opposite:
image

C-gyorfi (approved):

PR #33, opened 2022-08-04
Bicaws 2213 lock record when viewed

C-gyorfi:
### Changes
image


src/services/fetchAndTryLockCourtCase.ts
```diff
@@ -0,0 +1,41 @@
...
+): Promise<fetchAndTryLockCourtCaseResult> => {
+  const ok = await tryToLockCourtCase(dataSource, courtCaseId, currentUser.username)
+  if (!ok) {
+    return {
+      error: true
+    }
+  }
+
+  const courtCase = await getCourtCase(dataSource, courtCaseId, currentUser.visibleForces)
+  if (!courtCase) {
```
C-gyorfi, line 23: Is this a case of an error? `getCourtCase` would return null if the court case doesn't exist so I think we need to handle as not found

test/services/tryToLockCourtCase.integration.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    const result = await tryToLockCourtCase(dataSource, existingCourtCase.errorId, "Bichard01")
+    expect(isError(result)).toBe(false)
+
+    const record = await dataSource.getRepository(CourtCase).findOne({ where: { errorId: 0 } })
+    const actualCourtCase = record as CourtCase
+    expect(actualCourtCase.errorLockedById).toStrictEqual(anotherUser)
+    expect(actualCourtCase.triggerLockedById).toStrictEqual(anotherUser)
+  })
+
+  // These tests are out of scope until separate case and trigger locking is implemented
```
C-gyorfi, line 76: I think we want to remove these comments

test/services/fetchAndTryLockCourtCase.integration.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    await fetchAndTryLockCourtCase(
+      { username: "bichard01", visibleForces: ["36"] } as unknown as User,
+      unlockedCourtCase.error_id,
+      dataSource
+    )
+
+    const updatedCourtCase: CourtCase = await getCourtCase(dataSource, unlockedCourtCase.error_id, ["36"])
+    insertedCourtCase.errorLockedById = "bichard01"
+    insertedCourtCase.triggerLockedById = "bichard01"
```
C-gyorfi, line 72: This test is a bit confusing, we could remove `insertedCourtCase` and do 
```
expect(updatedCourtCase.errorLockedById).toStrictEqual("bichard01")
expect(updatedCourtCase.triggerLockedById).toStrictEqual("bichard01")
```
to run less queries

src/services/fetchAndTryLockCourtCase.ts
tomvaughan77: If you look at L23 in the case details page (`index.tsx`), this gets handled as a `notFound` already 😄

tomvaughan77 (approved):

PR #36, opened 2022-08-11
Refactor typeorm useage for tests

tomvaughan77:
General refactor of our data layer to make things a bit easier to work with.

* Typeorm is now used to insert testdata (removing lots of objects at look the same but are subtly different)
* Improved how the typeorm `dataSource` is gotten and used by the app and tests
* Removed xml -> json conversion for `aho` from our data layer. Now, we just pass around the xml as a string. If you want to use json, the page in question has to `parseAhoXml` it (and serialize it manually too)
* Added missing fields to `CourtCase` entity so we can stop using partials for everything
* Rejigged some of the folders in the `test` directory


cypress.config.ts
```diff
@@ -2,41 +2,46 @@
...
-        insertCourtCasesWithCourtNames(params: { courtNames: string[]; force: string }) {
-          return insertCourtCasesWithCourtNames(params.courtNames, params.force)
+        async insertCourtCasesWithCourtNames(params: { courtNames: string[]; force: string }) {
+          return await insertCourtCasesWithCourtNames(params.courtNames, params.force)
         },
 
-        insertCourtCasesWithDefendantNames(params: { defendantNames: string[]; force: string }) {
-          return insertCourtCasesWithDefendantNames(params.defendantNames, params.force)
+        async insertCourtCasesWithDefendantNames(params: { defendantNames: string[]; force: string }) {
+          return await insertCourtCasesWithDefendantNames(params.defendantNames, params.force)
```
emadkaramad, line 36: nitpick: I wonder if we need to have async/awaits for the tasks. I remember Cypress awaits tasks by default internally.

emadkaramad (approved):

cypress.config.ts
tomvaughan77: Whoops, saw this after I hit merge! 😬 Will see if I can remove them next time I'm in the area

PR #38, opened 2022-08-12
Bicaws 2219 add notes

emadkaramad:
PR to enable users to add note to a case

- Added `Add Note` page and feature
- Added `addNote` service
- Fixed lint errors and warnings
- Updated `getDummyCourtCase` to pass the data source to the function to fix `too many clients` errors in Cypress


test/util/insertCourtCases.ts
```diff
@@ -1,100 +1,94 @@
...
-import { InsertResult } from "typeorm"
+import { DataSource, InsertResult } from "typeorm"
 import getDataSource from "../../src/services/getDataSource"
 import DummyAho from "./error_list_aho.json"
 import DummyCourtCase from "./DummyCourtCase"
 
-const getDummyCourtCase = async (overrides?: Partial<CourtCase>): Promise<CourtCase> => {
-  const dataSource = await getDataSource()
-  return dataSource.getRepository(CourtCase).create({
+const getDummyCourtCase = (dataSource: DataSource, overrides?: Partial<CourtCase>): CourtCase =>
```
tomvaughan77, line 9: Just wondering why passing in `dataSource` is needed here? :)

src/features/AddNoteForm/AddNoteForm.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ComponentMeta, ComponentStory } from "@storybook/react"
+import { within } from "@storybook/testing-library"
+import AddNoteForm from "./AddNoteForm"
+
+export default {
+  title: "Features/CourtCase/Notes/Add",
+  component: AddNoteForm
+} as ComponentMeta<typeof AddNoteForm>
+
+export const WhenCaseNotLockedByAnotherUser: ComponentStory<typeof AddNoteForm> = () => (
```
tomvaughan77, line 11: Small thing - but you could write some extra tests here checking that you can actually write a note in, hit the `Add` button, and that it then is saved (rendered on screen correctly)?

tomvaughan77: Just feels a bit odd to me that a note-taking component doesn't have stories around it taking notes 📓

src/services/entities/CourtCase.ts
```diff
@@ -124,4 +124,11 @@ export default class CourtCase extends BaseEntity {
 
   @OneToMany(() => Note, (note) => note.courtCase)
   notes!: Relation<Note>[]
+
+  isLockedByAnotherUser(username: string) {
+    return (
+      (!!this.errorLockedById && this.errorLockedById !== username) ||
```
jaskaransarkaria, line 130: I think I've come across this before, in the db we actually use the username to lock the record and not the user id right? Is this something we need/ want to carry on doing?

jaskaransarkaria: or if not can `errorLockedById` -> `errorLockedByUsername`

jaskaransarkaria (approved):

tomvaughan77 (approved):

PR #39, opened 2022-08-16
Enable a user to resolve a trigger

monotypical:
This PR allows a user to resolve a trigger in the UI:

* Adds functionality and tests to resolve a trigger in `resolveTrigger.ts`, which sets the resolution columns on the trigger entity. If the trigger being resolved is the last unresolved trigger, it sets the resolution and status columns on the parent court case
* Adds a button to the case details screen which resolves a trigger through a form submission, and is only enabled when the trigger is not already resolved
* Adds the `ResolutionStatus` type, and uses this with a transformer for `error_status` and `trigger_status` columns on court cases, and the `status` column on triggers
* Adds an `optionalNullColumn` transformer which maps null database values to `undefined` in typescript, which makes handling optional values a bit neater


src/services/resolveTrigger.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (!updateTriggerSuccess) {
+        return updateTriggerSuccess
+      }
+
+      if (remainingUnresolvedTriggers === 0) {
+        const updateCaseResult = await entityManager
+          .createQueryBuilder()
+          .update(CourtCase)
+          .set({ triggerResolvedBy: resolver, triggerResolvedTimestamp: new Date(), triggerStatus: "Resolved" })
+          .where("error_id = :courtCaseId", { courtCaseId })
```
emadkaramad, line 59: minor: I wonder if we can use the alias instead of the actual table column name here? like `.createQueryBuilder("courtCase")...where("courtCase.errorId")`

emadkaramad (approved):

tomvaughan77 (approved):

tomvaughan77 (approved):

PR #40, opened 2022-08-18
Resubmit apply updates to hearing outcome

jaskaransarkaria:
This PR adds the base logic required to resubmit a case.

* adds logic to allow independent locking of triggers and exceptions
* add logic to amend properties on a case
* updated jest config so we don't have to import "expect" and "describe" at the top of every file

Outstanding steps include:

* ui integration (sending the amendments to the backend) - BICAWS-2272
* pushing the updated aho xml to the correct queue - BICAWS-2278

Outstanding TODOs:

* what if the user is amending multiple offences/ results currently this only handles single offence indexes and result indexes (we assume it would be best if we could do multiple) a trival task... (will be completed in a mop up ticket following this one)

---

*  cover the PNC update phase (waiting on core to implement this first)
*  set the status on the record (waiting to push the xml to the right queue) - BICAWS-2278
* unlock the record (waiting to push the xml to the right queue) - BICAWS-2278
* only amend the case if the case is unlocked or locked by the current user (verify this is the correct behaviour) - BICAWS-2272


bjpirt (approved):

src/services/updateCourtCase.ts
```diff
@@ -0,0 +1,25 @@
+import { DataSource, UpdateResult } from "typeorm/"
+import CourtCase from "./entities/CourtCase"
+
+const updateCourtCaseUpdatedHO = async (
```
C-gyorfi, line 4: The next function has the same name, but this function name doesn't correspond with the file name `updateCourtCase.ts`, did you forget to rename it?

src/services/getCourtCase.ts
```diff
@@ -1,15 +1,14 @@
...
   dataSource: DataSource | EntityManager,
-  courtCaseId: number,
-  forces: string[]
+  courtCaseId: number
 ): PromiseResult<CourtCase | null> => {
-  const courtCaseRepository = dataSource.getRepository(CourtCase)
-  const query = courtCasesByVisibleForcesQuery(courtCaseRepository, forces)
+  const repository = dataSource.getRepository(CourtCase)
+  const query = repository
+    .createQueryBuilder("courtCase")
```
C-gyorfi, line 11: Wouldn't we need to use the `courtCasesByVisibleForcesQuery` to ensure we show cases that are visible for the force?

PR #56, opened 2022-09-09
feat: 🎸 hearing outcome view

jaskaransarkaria:
First part of a 2 part PR.

This adds the hearing outcome needed for when we resubmit a case.

The second part will add the resubmission button and resubmit a case with no amendments


tsconfig.json
```diff
@@ -1,4 +1,10 @@
 {
+  "ts-node": {
```
jaskaransarkaria, line 2: This is needed to allow ts-node to run typescript scripts

cypress/tsconfig.json
```diff
@@ -3,6 +3,6 @@
   "compilerOptions": {
     "types": ["node", "cypress", "@testing-library/cypress"]
   },
-  "include": ["cypress/**/*", "../cypress.config.ts", "../node_modules/cypress"],
+  "include": ["cypress/**/*", "../cypress.config.ts", "../node_modules/cypress", "**/*.ts"],
```
jaskaransarkaria, line 6: Allows cypress scope to be limited to cypress files so the cypress test's expects, describes etc don't overwrite jests (now we don't have to explicitly import jests globals)

.eslintrc.json
```diff
@@ -29,20 +29,20 @@
...
+        "project": ["tsconfig.json", "cypress/tsconfig.json"]
       },
       "plugins": ["@typescript-eslint", "filenames"],
       "rules": {
         "curly": ["error", "all"],
         "no-console": "off",
         "no-plusplus": "off",
         "no-useless-escape": "off",
-        "require-await": "error",
+        "require-await": "off",
```
jaskaransarkaria, line 40: If you return an async fn you don't need to await it. eslint forces you to await it

src/components/HearingOutcome/HearingOutcome.tsx
```diff
@@ -0,0 +1,244 @@
+import { AnnotatedHearingOutcome } from "@moj-bichard7-developers/bichard7-next-core/build/src/types/AnnotatedHearingOutcome"
+import { Table, Tabs } from "govuk-react"
+import { useState } from "react"
+
+interface Props {
+  aho: AnnotatedHearingOutcome
+}
+
+const HearingOutcome: React.FC<Props> = ({ aho }) => {
+  const [currentTab, setCurrentTab] = useState(0)
```
tomvaughan77, line 10: We can't really start using `useState` yet until we're sure about IE versions (despite it being likey we can).

Any chance you can just take the `<Tabs>` out of this? Not too worried if the component is messy/unwieldy at this point, it'll get redesigned as part of the upcoming details page redesign.

tomvaughan77 (approved):

PR #58, opened 2022-09-16
Bicaws 2266 separate triggers and exceptions locking

emadkaramad:
This PR updates the locking logic to lock exceptions and triggers separately based on the user groups.

### DIFF
--- a/src/features/CourtCaseDetails/CourtCaseDetails.stories.tsx
+++ b/src/features/CourtCaseDetails/CourtCaseDetails.stories.tsx
@@ -31,7 +31,7 @@ const courtCase = {
 const aho = parseAhoXml(CourtCaseAho.hearingOutcomeXml) as AnnotatedHearingOutcome
 
 export const DetailsNotLockedByAnotherUser: ComponentStory<typeof CourtCaseDetails> = () => (
-  <CourtCaseDetails courtCase={courtCase} lockedByAnotherUser={false} aho={aho} />
+  <CourtCaseDetails courtCase={courtCase} lockedByAnotherUser={false} aho={aho} triggersVisible={true} />
 )
 
 DetailsNotLockedByAnotherUser.play = ({ canvasElement }) => {
@@ -48,7 +48,7 @@ DetailsNotLockedByAnotherUser.play = ({ canvasElement }) => {
 }
 
 export const DetailsLockedByAnotherUser: ComponentStory<typeof CourtCaseDetails> = () => (
-  <CourtCaseDetails courtCase={courtCase} lockedByAnotherUser={true} aho={aho} />
+  <CourtCaseDetails courtCase={courtCase} lockedByAnotherUser={true} aho={aho} triggersVisible={true} />
 )
 
 DetailsLockedByAnotherUser.play = ({ canvasElement }) => {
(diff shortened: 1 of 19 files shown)


src/features/CourtCaseDetails/CourtCaseDetails.stories.tsx
```diff
@@ -63,3 +63,22 @@ DetailsLockedByAnotherUser.play = ({ canvasElement }) => {
...
+
+export const TriggersNotVisibleToUser: ComponentStory<typeof CourtCaseDetails> = () => (
+  <CourtCaseDetails courtCase={courtCase} lockedByAnotherUser={false} aho={aho} triggersVisible={false} />
+)
+
+TriggersNotVisibleToUser.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  await expect(canvas.getByText(courtCase.ptiurn)).toBeInTheDocument()
+  await expect(canvas.getByText(courtCase.courtName)).toBeInTheDocument()
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
jaskaransarkaria, line 75: iirc I turned this eslint error off so I don't think we need this anymore

src/services/getCourtCase.ts
```diff
@@ -2,10 +2,7 @@ import { DataSource, EntityManager } from "typeorm"
 import CourtCase from "./entities/CourtCase"
 import PromiseResult from "../types/PromiseResult"
 
-const getCourtCase = async (
-  dataSource: DataSource | EntityManager,
-  courtCaseId: number
-): PromiseResult<CourtCase | null> => {
+const getCourtCase = (dataSource: DataSource | EntityManager, courtCaseId: number): PromiseResult<CourtCase | null> => {
```
jaskaransarkaria, line 5: How come the async is removed here 🤔

src/services/tryToLockCourtCase.ts
```diff
@@ -1,14 +1,16 @@
...
 
 const tryToLockCourtCase = (
   dataSource: DataSource,
   courtCaseId: number,
-  userName: string,
-  lockException = true,
-  lockTrigger = true
+  user: User
 ): Promise<UpdateResult | Error> | Error => {
+  const lockException = user.canLockExceptions
```
jaskaransarkaria, line 11: Can we just destructure this from user on line 9 so we don't need these variables?

src/services/unlockCourtCase.ts
```diff
@@ -1,19 +1,29 @@
 import { DataSource, UpdateResult } from "typeorm/"
+import { QueryDeepPartialEntity } from "typeorm/query-builder/QueryPartialEntity"
 import CourtCase from "./entities/CourtCase"
+import User from "./entities/User"
 
-const unlockCourtCase = async (dataSource: DataSource, courtCaseId: number): Promise<UpdateResult | Error> => {
+const unlockCourtCase = async (
+  dataSource: DataSource,
+  courtCaseId: number,
+  user: User
```
jaskaransarkaria, line 9: Same as above

src/services/unlockCourtCase.ts
```diff
@@ -1,19 +1,29 @@
...
+    if (user.canLockExceptions) {
+      setFields.errorLockedByUsername = null
+    }
+    if (user.canLockTriggers) {
+      setFields.triggerLockedByUsername = null
+    }
+
+    query.set(setFields).where("error_id = :id", { id: courtCaseId })
+
+    return await query.execute()
```
jaskaransarkaria, line 26: I don't think we need the await when you're returning it

src/services/resubmitCourtCase/index.ts
```diff
@@ -1 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
```
jaskaransarkaria, line 1: I thought I set this to warning too 🤔

jaskaransarkaria (approved):

src/features/CourtCaseDetails/CourtCaseDetails.stories.tsx
emadkaramad: Good spot. 👍

src/services/unlockCourtCase.ts
emadkaramad: Good shout! I also remove the try/catch and use `catch()` function instead.

src/services/resubmitCourtCase/index.ts
emadkaramad: Not sure tbh. My concern with warnings is that it is easy to miss them. We might end up in an inconsistent situation where some modules are exported as default and some are not.

PR #75, opened 2022-10-06
Allow displaying cases with annotated message in PNCUpdateDataset format

C-gyorfi:


test/util/DummyCourtCase.ts
```diff
@@ -9,7 +9,7 @@ const DummyCourtCase: Partial<CourtCase> = {
   triggerReason: "",
   errorId: 79507,
   messageId: "cf60600b-52a0-4bc7-b117-3acfb851a7f5",
-  phase: 2,
+  phase: 1,
```
C-gyorfi, line 12: Updated the default stubbed message to phase 1 because we are using AHO XML format in the annotated message, phase 2 should use `PNCUpdateDataset` XML format

jaskaransarkaria (approved):

PR #95, opened 2022-11-02
Add sorting and filtering for urgent cases on the case listing screen

monotypical:
This PR:
* Makes the "Urgent" column in the case listing screen sortable
* Adds a checkbox for filtering cases by urgency on the case listing screen
* Adds an option to `listCourtCases` to only retrieve urgent cases
* Adds cypress tests for both sorting and filtering urgent cases


src/services/listCourtCases.ts
```diff
@@ -33,6 +33,10 @@ const listCourtCases = async (
     query.andWhere("courtCase.errorCount > 0")
   }
 
+  if (urgentFilter) {
```
tomvaughan77, line 36: Mind just chucking an integration test in for this functionality?

monotypical: Added in [10af358](https://github.com/ministryofjustice/bichard7-next-ui/pull/95/commits/10af3582ecc0119d508095fef861b5f89786fb60)

tomvaughan77 (approved):

PR #106, opened 2022-11-09
Filter court cases by named date ranges

C-gyorfi:
### Screenshots
When the date range "Yesterday" is selected:
Screenshot 2022-11-16 at 14 40 28

When no date range is selected:
Screenshot 2022-11-16 at 14 41 21

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -17,6 +17,7 @@ import type CourtCase from "services/entities/CourtCase"
 import { Heading } from "govuk-react"
 import CourtCaseWrapper from "features/CourtCaseFilters/CourtCaseFilterWrapper"
 import { mapDateRange, validateNamedDateRange } from "utils/validators/validateDateRanges"
+import { validateQueryParams } from "utils/validators/validateQueryParams"
 
 interface Props {
   user: User
@@ -30,7 +31,6 @@ interface Props {
   pageNum: number
 }
 
-const validateQueryParams = (param: string | string[] | undefined): param is string => typeof param === "string"
 const validateOrder = (param: unknown): param is QueryOrder => param === "asc" || param == "desc" || param === undefined
 const validCourtCaseTypes = ["Triggers", "Exceptions"]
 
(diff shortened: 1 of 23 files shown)


emadkaramad (approved):

src/pages/index.tsx
```diff
@@ -25,24 +27,24 @@ interface Props {
...
 const validCourtCaseTypes = ["Triggers", "Exceptions"]
 
 export const getServerSideProps = withMultipleServerSideProps(
   withAuthentication,
   async (context: GetServerSidePropsContext<ParsedUrlQuery>): Promise<GetServerSidePropsResult<Props>> => {
     const { currentUser, query } = context as AuthenticationServerSidePropsContext
-    const { orderBy, pageNum, type, keywords, maxPageItems, order, urgency } = query
+    const { orderBy, pageNum, type, keywords, maxPageItems, order, urgency, dateRange } = query
     const courtCaseTypes = [type].flat().filter((t) => validCourtCaseTypes.includes(String(t))) as Filter[]
     const validatedMaxPageItems = validateQueryParams(maxPageItems) ? maxPageItems : "5"
```
emadkaramad, line 43: Can we not simplify this to `maxPageItems || "5"`?
`validateQueryParams` checks if the type is a string but I think all values we get from query string are type of string anyway.
The same question for other fields below this one.

monotypical: > I think all values we get from query string are type of string anyway.

Not quite, if the url parameter is not present you get undefined, which `maxPageItems || "5"` would work fine for, but if the URL parameter is set multiple times, for instance in the url `/bichard?maxPageItems=1&maxPageItems=2`, Next gives you an array of all values, in this example the array `["1", "2"]`.

We won't ever set these parameters to multiple values, but users could set the URL parameters to whatever they want, so we want to revert back to the default of `"5"` when the parameter is set multiple times. Either way we need to handle the value being  `string[]` type, since the type of values in `router.query` is `string | string[] | undefined`

We are planning to support multiple date ranges when we implement the SLA checkboxes, so this will need some more thought when we do that ticket, but we'll cross that bridge when we get there

emadkaramad: Thanks Alice. I didn't know NextJS query also parse to `string[]`. That makes sense. 👍

PR #108, opened 2022-11-10
refactor: 💡 resubmit fn

jaskaransarkaria:
resubmit xml now matches the legacy resubmit functionality

* Pulls in new `convertAhoToXml` fn and turns off validations
* correctly updates the `user_updated_flag` in db
* fixes a bug in a no update resubmit (amendments were parsed incorrectly)
* reorder params so `dataSource` is consistent across fns


src/services/resubmitCourtCase/resubmitCourtCase.ts
```diff
@@ -37,26 +41,41 @@ const resubmitCourtCase = async (
...
+
+        return amendedCourtCase
       }
+    )
+
+    if (isError(courtCase)) {
+      throw courtCase
+    }
+
+    const generatedXml = convertAhoToXml(courtCase, false)
```
jaskaransarkaria, line 71: Turn off validations to create the correct xml

src/services/updateCourtCaseAho.ts
```diff
@@ -13,8 +14,8 @@ const updateCourtCaseAho = async (
       .createQueryBuilder()
       .update(CourtCase)
       .set({
-        hearingOutcome: updatedHo,
-        userUpdatedFlag: 1
+        updatedHearingOutcome: updatedHo,
```
jaskaransarkaria, line 17: update the `updated_msg` column

src/services/updateCourtCaseAho.ts
```diff
@@ -13,8 +14,8 @@ const updateCourtCaseAho = async (
...
+        userUpdatedFlag: userUpdated ? 1 : 0
```
jaskaransarkaria, line 18: Only update this flag if there are actually updates

src/services/resubmitCourtCase/resubmitCourtCase.ts
```diff
@@ -18,14 +20,16 @@ const resubmitCourtCase = async (
   currentUser: User
 ): PromiseResult<AnnotatedHearingOutcome | Error> => {
   try {
-    return await dataSource.transaction("SERIALIZABLE", async (entityManager) => {
-      const lockResult = await tryToLockCourtCase(entityManager, +courtCaseId, currentUser)
+    const lockResult = await tryToLockCourtCase(dataSource, +courtCaseId, currentUser)
```
jaskaransarkaria, line 23: This should be outside of the db transaction

src/components/EditableField/EditableField.tsx
```diff
@@ -10,7 +10,7 @@ const EditableField: React.FC<{
   amendFn: (newValue: string) => void
 }> = ({ aho, objPath, amendFn }) => {
   const result: ExceptionCode | null = isException(aho, objPath)
-  const value = get(aho, objPath, "").toString()
+  const value = (get(aho, objPath, "") ?? "").toString() // object is returned for dates
```
jaskaransarkaria, line 13: because an object cannot be rendered in react we need to transform it to string but we can't use `get(aho, objPath, "")?.toString()` as that could render "undefined".

emadkaramad (approved):

PR #111, opened 2022-11-14
Added test for court case with updated court offence sequence number

joe-fol:
- Users can now resubmit updates to multiple offences and results within offences whilst maintaining functionality to update simpler fields (relevant amend fns updated to take an array of updates and tests added)
- fixes a bug which was adding new updates per key stroke
- refactors editable field component to so it is the triggers the update events and also displays ResultVariableText field
- refactors the HearingOutcome component
- adds e2e test for resubmitting and updating


cypress/e2e/court-cases/caseDetails.cy.ts
```diff
@@ -145,10 +146,10 @@ describe("Case details", () => {
       })
     })
 
-    it("should redirect to the user-service if the auth token provided is for a non-existent user", () => {
+    it.skip("should redirect to the user-service if the auth token provided is for a non-existent user", () => {
```
jaskaransarkaria, line 149: This has been failing before this pr so I've skipped it and added a ticket https://dsdmoj.atlassian.net/browse/BICAWS-2540

cypress/e2e/court-cases/caseDetails.cy.ts
```diff
@@ -49,10 +50,10 @@ describe("Case details", () => {
 
       cy.injectAxe()
 
-      cy.checkA11y(undefined, a11yConfig, logAccessibilityViolations)
+      cy.checkA11y("#content", a11yConfig, logAccessibilityViolations)
```
jaskaransarkaria, line 53: This test was continually failing as it was returning empty html when defined when it should grab the whole page, this gets the <Main /> element and asserts on that explicitly.

src/services/amendCourtCase/applyAmendmentsToAho.test.ts
```diff
@@ -92,17 +92,19 @@ describe("applyAmendmentsToAho", () => {
...
+        }
+      ]
     } as Amendments
 
     aho.AnnotatedHearingOutcome.HearingOutcome.Case.HearingDefendant.Offence = [
-      dummyOffence,
-      dummyOffence,
-      dummyOffence,
-      dummyOffence
+      cloneDeep(dummyOffence),
```
jaskaransarkaria, line 104: We have to use `cloneDeep()` to avoid problems with deeply nested objects values being passed by reference

cypress/e2e/court-cases/caseDetails.cy.ts
C-gyorfi: I think we want to run accessibility tests on the entire page in e2e tests, not just this element. Bear in mind accessibility tests are failing in dev mode because NEXT doesn't render some of the landmarks correctly. You can try npm run build && npm run start which should pass these tests

cypress/e2e/court-cases/caseDetails.cy.ts
```diff
@@ -145,10 +146,10 @@ describe("Case details", () => {
       })
     })
 
-    it("should redirect to the user-service if the auth token provided is for a non-existent user", () => {
+    it.skip("should redirect to the user-service if the auth token provided is for a non-existent user", () => {
       cy.login("bichard01@example.com", "password")
       cy.clearCookies()
-      cy.visit("/bichard/court-cases/0")
+      cy.visit("/bichard/court-cases/0", { failOnStatusCode: false })
```
C-gyorfi, line 152: So this change hasn't fixed the test? If so I think it make sense to skip it, its a test that is known to be flaky

src/components/HearingOutcome/utils.ts
```diff
@@ -0,0 +1,52 @@
+import { AmendmentArrValues, AmendmentValues, IndividualAmendmentArrValues } from "types/Amendments"
+
+const isAmendmentArrValue = (value: AmendmentValues, indexToUpdate: number): value is AmendmentArrValues =>
+  (value as AmendmentArrValues)[indexToUpdate].updatedValue !== undefined
+
+export const isAmendmentArr = (value: AmendmentValues): value is AmendmentArrValues => typeof value !== "string"
```
C-gyorfi, line 6:
```suggestion
export const isAmendmentArr = (value: AmendmentValues): value is AmendmentArrValues => Array.isArray(value)
```

Would this be more explicit?

test/services/resubmitCourtCase.integration.test.ts
```diff
@@ -33,6 +35,7 @@ describe("resubmit court case", () => {
   })
 
   it("should resubmit a court case with no updates", async () => {
+    // set up court case in the right format to insert into the db
```
C-gyorfi, line 38: Do we need these comments?

test/utils/getCourtCaseById.ts
```diff
@@ -0,0 +1,10 @@
+import CourtCase from "../../src/services/entities/CourtCase"
+import getDataSource from "../../src/services/getDataSource"
+
+export const getCourtCaseById = async (caseId: number): Promise<CourtCase> => {
```
C-gyorfi, line 4: There is a `getCourtCase` function that is doing something very similar, I wonder if we could modify/reuse that

C-gyorfi (approved):

PR #113, opened 2022-11-18
Add signout behaviour to header bar in New-UI

donnyhyon:
Added `Sign out` behaviour to header bar. 
Copied from user service.


src/components/Header/Header.tsx
```diff
@@ -42,7 +42,7 @@ const Header: React.FC<HeaderProps> = ({ organisationName, serviceName, userName
                 </li>
                 <li className="moj-header__navigation-item">
                   {/* TODO: Link to userService */}
```
monotypical, line 44: We can get rid of this TODO now as well

donnyhyon: Thanks! Have now updated.

PR #114, opened 2022-11-18
Run e2e tests in CI

C-gyorfi:


.circleci/config.yml
```diff
@@ -171,7 +171,7 @@ jobs:
           name: Update Chromatic baseline
           command: npm run chromatic:ci:main
 
-  build_and_e2e_test:
+  build_and_cypress_test:
```
monotypical, line 174: FYI I think we will need to change settings on this repo so that the `build_and_e2e_test` required status gets updated, otherwise we won't be able to merge PRs while it waits for the status to be reported

monotypical (approved):
Nice work Csaba! 🙌

PR #120, opened 2022-11-23
Filter cases by locked status

monotypical:
In this PR:
* Added UI to select a locked status filter
* Added cypress tests for filtering cases in the case listing screen by their locked status
* Added ability for `listCourtCases` to take an optional locked filter
* Add unit tests for new `listCourtCases` filtering
* Added `insertMultipleDummyCourtCasesWithLock` helper for use in tests
* Made case filter labels more consistent

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -29,6 +29,7 @@ interface Props {
   totalPages: number
   dateRange: string | null
   pageNum: number
+  lockedFilter: string | null
 }
 
 const validateOrder = (param: unknown): param is QueryOrder => param === "asc" || param == "desc" || param === undefined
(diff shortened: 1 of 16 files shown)


src/types/CaseListQueryParams.ts
```diff
@@ -14,11 +14,12 @@ export type Urgency = "Urgent" | "Non-urgent" | undefined
 export type CaseListQueryParams = {
   orderBy?: string
   order?: QueryOrder
-  resultFilter?: Filter[]
+  resultFilter?: TriggerExceptionFilter[]
```
C-gyorfi, line 17: I like that you renamed this, but I wonder if we could use a term that also used in the design, maybe:
```suggestion
  reason?: Reason[]
```

src/types/CaseListQueryParams.ts
```diff
@@ -14,11 +14,12 @@ export type Urgency = "Urgent" | "Non-urgent" | undefined
...
   defendantName?: string
   urgentFilter?: Urgency
   forces: string[]
   pageNum?: string
   maxPageItems: string
   courtDateRange?: CourtDateRange
+  lockedFilter?: boolean
```
C-gyorfi, line 24: `Filter` seems redundant in the attribute name

src/pages/index.tsx
```diff
@@ -94,7 +102,8 @@ const Home: NextPage<Props> = ({
   courtCaseTypes,
   keywords,
   dateRange,
-  urgentFilter
+  urgentFilter,
```
C-gyorfi, line 105: Could we use `urgent` and `locked` here?

C-gyorfi (approved):
Awesome work 🚀 I left some naming suggestions but looks good!

src/types/CaseListQueryParams.ts
donnyhyon: Thanks Csaba! Have made these changes now.

PR #130, opened 2023-01-04
NewUI: fixed Page component padding

donnyhyon:
This PR removes the white spacing on either side of the page. 
Created a custom Page component `PageTemplate` as the Gov-react component enforced a set width and was complex to change. 

- Used JSS to apply custom styling. 
- Made a central JSS file to store all custom styles.

Note:
A `<h1>` header is required to pass A11y tests which is not included in the figma designs. Need feedback from designers on this. This may need modified in future. 
https://www.figma.com/file/LHxGJpfSjuTczcqrFzUQNc/06_B7_22-Handover-hifi-case-list-screens?node-id=321%3A2571&t=5X7f8vkVtEOL1YGs-0

Before
image

After
image

image

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,13 +1,24 @@
+import { createUseStyles } from "react-jss"
+
 interface HeaderProps {
   serviceName: string
   organisationName: string
   userName: string
 }
+
+const useStyles = createUseStyles({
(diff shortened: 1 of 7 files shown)


monotypical (approved):
This looks a lot nicer now that the content has more room

PR #133, opened 2023-01-10
Filter by case state

C-gyorfi:
Add case state filter options according to design
- `Unresolved & resolved cases` or `Resolved cases`
- `Unresolved cases` returned if no case state filter is applied

image


src/features/CourtCaseFilters/CourtCaseFilter.tsx
```diff
@@ -48,9 +55,12 @@ const reducer = (state: Filter, action: FilterAction): Filter => {
     } else if (action.type === "date") {
       newState.dateFilter.value = undefined
       newState.dateFilter.label = undefined
+    } else if (action.type === "caseState") {
+      newState.caseStateFilter.value = undefined
+      newState.caseStateFilter.label = undefined
     } else if (action.type === "locked") {
       newState.lockedFilter.value = undefined
-      newState.dateFilter.label = undefined
+      newState.lockedFilter.label = undefined
```
monotypical, line 63: good catch!

monotypical (approved):
Nice work Csaba!

PR #148, opened 2023-01-17
Allow searching by keywords

C-gyorfi:
### Context
As a user, I’d like to be able to filter cases by keywords in all of the following fields `Defendent name`, `Court name`, `Reason`, `PTIURN`

### Changes:
- Update `listCourtCases` to take defendant name, court name, PTIURN and reason as query params
  - Reason search utilises the`trigger_reason` column (includes all trigger codes with comma separation) and `error_report` column (just like [legacy B7](https://github.com/ministryofjustice/bichard7-next/blob/main/bichard-backend/src/main/java/uk/gov/ocjr/mtu/br7/errorlistmanager/data/RecordSelectionSQLGenerator.java#L1276), which includes all error codes and field names with comma separation)
 - Add search fields and implement filter chips 

### Before
image

### After
image


cypress/e2e/filter/filterCases.cy.ts
```diff
@@ -136,6 +136,99 @@ describe("Case list", () => {
...
+        force: "011111"
+      })
+
+      cy.visit("/bichard")
+
+      cy.get("button[id=filter-button]").click()
+      cy.get("input[id=court-name]").type("Manchester Court")
+
+      cy.get("button[id=search]").click()
+      cy.get("tr").not(":first").get("td:nth-child(3)").contains("Manchester Court")
```
agnieszkazbikowska-madetech, line 151: I'm wondering if it would be better to have `contains` on its own. Are the rows so important here? It would be easier to read and also if we add a new line in the future, it wouldn't break all the tests.

cypress/e2e/filter/filterCases.cy.ts
```diff
@@ -136,6 +136,99 @@ describe("Case list", () => {
...
+      cy.get("tr").not(":first").get("td:nth-child(3)").contains("London Court").should("not.exist")
+      cy.get("tr").not(":first").get("td:nth-child(3)").contains("Bristol Court").should("not.exist")
+      cy.get("tr").should("have.length", 2)
+      cy.get(".moj-filter-tags a.moj-filter__tag").contains("Manchester Court")
+
+      // Removing filter tag
+      cy.get(".moj-filter-tags a.moj-filter__tag").contains("Manchester Court").click({ force: true })
```
agnieszkazbikowska-madetech, line 158: This is used a few times across the file. We could create a function for this that would take an argument (`Manchester Court`) in this case.

src/features/CourtCaseFilters/FilterChipSection.tsx
```diff
@@ -113,6 +94,62 @@ const FilterChipSection: React.FC<Props> = ({ state, dispatch, sectionState, mar
             ))}
         </ul>
       </If>
+
+      <FilterChipRow
```
monotypical, line 98: These look nicer being extracted into a component

monotypical (approved):

cypress/e2e/filter/filterCases.cy.ts
C-gyorfi: I agree. I think we can do a follow-up PR in which we can refactor all the filter tests! 🚀

PR #160, opened 2023-01-26
Show whether cases are locked in the case list

monotypical:
This PR primarily adds two columns to the case list page:

- The untitled, leftmost column which displays a lock icon if the case is locked
- The "Locked By" column on the far right, which displays the username of the user who has the case locked as a tag

This PR also:

- adds cypress and storybook tests for this functionality
- extracts colour definitions into their own file
- adds the lock icon from BoxIcons (MIT licensed so no attribution needed https://boxicons.com/usage#license)
- prevents the accessibility tests from running before the page has fully loaded, which sometimes happens when running locally and results in a test failure

Screenshot:
Screenshot 2023-01-26 at 16 45 37

Design: https://www.figma.com/file/LHxGJpfSjuTczcqrFzUQNc/06_B7_22-Handover-hifi-case-list-screens?node-id=327%3A5000&t=9PpSCC2YKzW7V0jz-0


cypress/e2e/court-cases/notes/add.cy.ts
```diff
@@ -46,6 +46,9 @@ describe("Case details", () => {
 
       cy.injectAxe()
 
+      // Wait for the page to fully load
+      cy.get("h1")
```
C-gyorfi, line 50: Awesome! 🥇

C-gyorfi (approved):
💯

PR #161, opened 2023-01-26
Bicaws 2608 Add reason field to display error and trigger details

C-gyorfi:
### Context
Currently, we show Triggers and Exceptions in two separate columns; these should be displayed under the reason column

### Changes
- Map the errors from `errorReport` field and display according to design
- Install the standing data package
- Utilise the `trigger-definitions` from the standing data package to append the triggers with a short description
- Update the order by logic in `listCourtCases` so it uses both `errorReason` and `triggerReason` fields when ordering by reason(since reasons live in two separate columns in DB this is the best sorting mechanism I could come up with)
- Fix a bug when searching for cases by trigger code

image

### DIFF
--- a/cypress/e2e/index.cy.ts
+++ b/cypress/e2e/index.cy.ts
@@ -60,7 +60,7 @@ describe("Case list", () => {
 
         // Wait for the page to fully load
         cy.get("h1")
-        
+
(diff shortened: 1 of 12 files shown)


monotypical (approved):
🎉

PR #165, opened 2023-02-02
Unlock cases from case list

C-gyorfi:
### Context
As a supervisor, I would like to unlock a case by clicking on the locked by user name.

### Changes
- Update `unlockCourtCase` logic:
  - to allow unlocking triggers or exceptions
  - to allow supervisor users to unlock a case that is locked by another user
- Unlock exceptions by sending a post request to the case list page with params `/?unlockException=:caseID`
- Unlock triggers by sending a post request to the case list page with params `/?unlockTrigger=:caseID`
  - This should be refactored in the future using an API endpoint(B7 API)

image

### DIFF
--- a/src/services/queries/courtCasesByVisibleForcesQuery.ts
+++ b/src/services/queries/courtCasesByVisibleForcesQuery.ts
@@ -1,13 +1,10 @@
 import CourtCase from "../entities/CourtCase"
-import { Brackets, In, Repository, SelectQueryBuilder } from "typeorm"
-import KeyValuePair from "../../types/KeyValuePair"
+import { Brackets, In, Like, SelectQueryBuilder, UpdateQueryBuilder } from "typeorm"
 
 const courtCasesByVisibleForcesQuery = (
-  repository: Repository<CourtCase>,
+  query: SelectQueryBuilder<CourtCase> | UpdateQueryBuilder<CourtCase>,
   forces: string[]
-): SelectQueryBuilder<CourtCase> => {
-  const query = repository.createQueryBuilder("courtCase")
-
+): SelectQueryBuilder<CourtCase> | UpdateQueryBuilder<CourtCase> => {
   query.where(
     new Brackets((qb) => {
       if (forces.length < 1) {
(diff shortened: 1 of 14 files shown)


src/services/unlockCourtCase.ts
```diff
@@ -6,27 +6,39 @@ import User from "./entities/User"
...
   }
-  if (canLockTriggers) {
+  if (shouldUnlockTriggers) {
     setFields.triggerLockedByUsername = null
   }
 
   query.set(setFields).where("error_id = :id", { id: courtCaseId })
 
-  return query.execute().catch((error: Error) => error)
+  if (!isSupervisor && shouldUnlockExceptions) {
```
monotypical, line 33: Currently, supervisors can unlock any court case given in the query parameters to the case list page. I wonder if we should prevent supervisors from unlocking cases from other forces?

monotypical (commented):
Great work Csaba! 🎉

src/services/unlockCourtCase.ts
C-gyorfi: I think its low risk since they wouldn't see those cases anyway, but I refactored to make sure its absolutely impossible:
[6e94a3e](https://github.com/ministryofjustice/bichard7-next-ui/pull/165/commits/6e94a3e2261e5b17ab8a90793df77c14ce86b798)

src/services/queries/courtCasesByVisibleForcesQuery.ts
```diff
@@ -1,28 +1,24 @@
...
       }
 
-      forces.forEach((force, i) => {
-        const args: KeyValuePair<string, string> = {}
-        args[`force${i}`] = force
-        // use different named parameters for each force
+      forces.forEach((force) => {
         if (force.length === 1) {
-          qb.orWhere(`courtCase.orgForPoliceFilter like :force${i} || '__%'`, args)
+          const condition = { orgForPoliceFilter: Like(`${force}__%`) }
```
monotypical, line 17: This is a lot nicer this way

PR #172, opened 2023-02-10
Bug fix: multiple search fields

C-gyorfi:
Using `defendantName` and `courtName` searches broke the search because both fields used the ambiguous `:name` param Updated query and added test

### DIFF
--- a/cypress/e2e/filter/filterCases.cy.ts
+++ b/cypress/e2e/filter/filterCases.cy.ts
@@ -230,7 +230,7 @@ describe("Case list", () => {
       confirmMultipleFieldsDisplayed(["Case00000", "Case00001", "Case00002"])
     })
 
-    it("Should let to use all search fields", () => {
+    it("Should let users use all search fields", () => {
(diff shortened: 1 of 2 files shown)


cypress/e2e/filter/filterCases.cy.ts
```diff
@@ -230,6 +230,54 @@ describe("Case list", () => {
       confirmMultipleFieldsDisplayed(["Case00000", "Case00001", "Case00002"])
     })
 
+    it("Should let to use all search fields", () => {
```
monotypical, line 233: typo here

monotypical (approved):
good catch!

cypress/e2e/filter/filterCases.cy.ts
C-gyorfi: Fixed [6a49834](https://github.com/ministryofjustice/bichard7-next-ui/pull/172/commits/6a498340c20f60a19d906f55e660c6bea52e71ee)

monotypical (approved):

PR #179, opened 2023-02-16
Bicaws 2617 display sorting column order

joe-fol:


cypress/component/ColumnOrderIcons.cy.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react"
+import ColumnOrderIcons from "../../src/features/CourtCaseFilters/ColumnOrderIcons"
+
+describe("ColumnOrderIcons.cy.tsx", () => {
```
monotypical, line 4: these cypress component tests are great to see 🤩

cypress/component/ColumnOrderIcons.cy.tsx
```diff
@@ -0,0 +1,14 @@
...
+  it("up arrow", () => {
+    cy.mount(<ColumnOrderIcons columnName={"defendantName"} currentOrder={"asc"} orderBy={"defendantName"} />)
```
monotypical, line 6: could we add some assertions here that an element with the class `upArrow` exists? and likewise for the other tests in this file, then we're checking that the right arrows are showing

cypress/e2e/index.cy.ts
```diff
@@ -178,6 +178,13 @@ describe("Case list", () => {
         cy.visit("/bichard")
 
         cy.findByText("Court Name").click()
+        cy.findByText("Court Name").parent().siblings().get('*[class^="icon"]').should("exist")
+        cy.findByText("Court Name")
+          .parent()
+          .siblings()
+          .get('*[class^="icon"]')
+          .get('*[class^="upArrow"]')
```
monotypical, line 186: since the `upArrow` and `downArrow` classes aren't generated by `react-jss` and are static, we can shorten this to `.upArrow` instead of `*[class^="upArrow"]`

monotypical (commented):
Great work 🎉

I think it would be nice to have some shortcuts to run the cypress component tests from `package.json` as something like `cypress:run:component`, and also to run these in the CI, as currently we only run the e2e cypress tests

We need to pass `--component` to cypress to run the component tests https://docs.cypress.io/guides/guides/command-line

monotypical (approved):

PR #181, opened 2023-02-16
BICAWS-2506: Tech Debt: Resolve accessibility issue with conditional fields

agnieszkazbikowska-madetech:


monotypical (approved):

src/components/CourtDateFilter/CourtDateFilterOptions.tsx
```diff
@@ -43,7 +43,6 @@ const CourtDateFilterOptions: React.FC<Props> = ({ dateRange, dispatch, customDa
         <RadioButton
           name={"courtDate"}
           id={"date-range"}
-          dataAriaControls={"conditional-date-range"}
```
C-gyorfi, line 46: Unfortunatelly if we remove this, it breaks the collapsable radio buttons(they are aways displayed):
image

cypress/support/helpers.ts
```diff
@@ -0,0 +1,4 @@
+export const waitForPageToLoad = () => {
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(1000)
```
C-gyorfi, line 3: This will always add a second to the test run whenever we call it. Could we use the following instead?
```
cy.get("h1", {timeout: 1000})
```

agnieszkazbikowska-madetech: We had this before and the tests were failing. We've had a look with @monotypical .

monotypical: Yeah, unfortunately with just waiting for h1 we would still sometimes get accessability violations with the tests being run before the page had fully loaded

PR #193, opened 2023-02-27
BICAWS-2713: Bug: Changing court date selection doesn't change selected filters

agnieszkazbikowska-madetech:
* add tests to replicate the behaviour in the ticket
* fix types for `customDateFrom` and `customDateTo`
* fix the bug
* create `getCustomDateRangeLabel()` helper function
* create `filterByDateRange()` and `filterByCustomDateRange()` helper functions for cypress
* replace `defaultValue` with `value` for `DateInput`

### DIFF
--- a/cypress/e2e/filter/filterChips.cy.ts
+++ b/cypress/e2e/filter/filterChips.cy.ts
@@ -6,7 +6,7 @@ export function removeFilterChip() {
   cy.get(".moj-filter__tag").should("not.exist")
 }
 
-function filterByDateRange(dateRangeId: string) {
+export function filterByDateRange(dateRangeId: string) {
   cy.get("#date-range").click()
   cy.get(dateRangeId).click()
 }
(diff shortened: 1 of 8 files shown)


cypress/e2e/filter/filterChips.cy.ts
```diff
@@ -6,6 +6,17 @@ export function removeFilterChip() {
   cy.get(".moj-filter__tag").should("not.exist")
 }
 
+export function filterByDateRange(dateRangeId: string) {
+  cy.get("#date-range").click()
+  cy.get(dateRangeId).click()
+}
+
+function filterByCustomDateRange(dateFrom: string, dateTo: string) {
```
C-gyorfi, line 14: I think in most cases, we prefer using arrow functions over "traditional functions", while in some cases, it just a style choice. For instance I like arrow functions because they have a more concise syntax and quite readable. But there are technical differences too:
- const prevents reassignment but `function` not
- arrow functions has to be declared before calling it, while functions can be defined after calling them

nit picking is over :)

agnieszkazbikowska-madetech: It makes sense. I've updated!

PR #195, opened 2023-03-01
Update case list title to match design

C-gyorfi:
### Screenshot
image

### DIFF
--- a/cypress/e2e/index.cy.ts
+++ b/cypress/e2e/index.cy.ts
@@ -287,7 +287,7 @@ describe("Case list", () => {
       cy.get("tr").not(":first").eq(2).get("td:nth-child(7)").contains(`3`).should("exist")
     })
 
-    it.only("should be able to navigate to the case details page and back", () => {
+    it("should be able to navigate to the case details page and back", () => {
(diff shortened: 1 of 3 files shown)


cypress/e2e/index.cy.ts
```diff
@@ -287,7 +287,7 @@ describe("Case list", () => {
       cy.get("tr").not(":first").eq(2).get("td:nth-child(7)").contains(`3`).should("exist")
     })
 
-    it("should be able to navigate to the case details page and back", () => {
+    it.only("should be able to navigate to the case details page and back", () => {
```
agnieszkazbikowska-madetech, line 290: Do we want to keep `only`?

agnieszkazbikowska-madetech (approved):

cypress/e2e/index.cy.ts
C-gyorfi: no :D

C-gyorfi, line 290:
```suggestion
    it("should be able to navigate to the case details page and back", () => {
```

PR #199, opened 2023-03-03
BICAWS-2672: Add a confirmation step and show when a case is unlocked

agnieszkazbikowska-madetech:
**BICAWS-2672 - Add a confirmation step and show when a case is unlocked**

1. create `CaseUnlockedTag` component,
2. create `UnlockConfirmation` component,
3. create `LockedByButton` component,
4. update `LockedByTag` component,
5. create a custom `button--tag` class,
6. create `LockedByText` component,
7. create a `CaseUnlocked` story,
8. update e2e tests for the unlocking case functionality,
9. create `unlockCase` cy helper function,
10. create unit tests for `LockedByText` and `LockedByButton`,
11. create new colour variables,


src/utils/colours.ts
```diff
@@ -1,6 +1,10 @@
 const darkGrey = "#62696D"
+const gdsLightGrey = "#f3f2f1"
+
 const tagBlue = "#e9f1f8"
 const textBlue = "#114e81"
-const gdsLightGrey = "#f3f2f1"
+const navy = "#1D70B8"
```
C-gyorfi, line 6: I think it's nice that we extracted the blue colour, but maybe we could follow the gov UK naming, which is blue? https://design-system.service.gov.uk/styles/colour/

agnieszkazbikowska-madetech:
> Looks good, there is more thing that's on the design: image
> 
> Although, I'm unsure what the behaviour of this is, how long we would display this success badge, maybe a question for designers?

@C-gyorfi this is a very good question. I'll ask in slack. I assumed that this stays there for ever.

src/utils/colours.ts
agnieszkazbikowska-madetech: Thanks for sharing the link. I'll rename `navy` to `blue`.

src/features/CourtCaseList/CourtCaseListEntry.tsx
```diff
@@ -156,6 +165,7 @@ const CourtCaseListEntry: React.FC<Props> = ({ courtCase, currentUser }: Props)
           ) : (
             <LockedByTag lockedBy={errorLockedByUsername} />
           )}
+          {<CaseUnlockedTag isCaseUnlocked={exceptionHasBeenRecentlyUnlocked && triggerHasBeenRecentlyUnlocked} />}
```
C-gyorfi, line 168: I think we'd need check whether the case actually unlocked, something like:
```suggestion
          {<CaseUnlockedTag isCaseUnlocked={exceptionHasBeenRecentlyUnlocked && triggerHasBeenRecentlyUnlocked && !errorLockedByUsername} />}
```

Otherwise users may see something like this:
image

src/features/CourtCaseList/CourtCaseListEntry.tsx
```diff
@@ -107,6 +115,7 @@ const CourtCaseListEntry: React.FC<Props> = ({ courtCase, currentUser }: Props)
           ) : (
             <LockedByTag lockedBy={triggerLockedByUsername} />
           )}
+          {<CaseUnlockedTag isCaseUnlocked={!exceptionHasBeenRecentlyUnlocked && triggerHasBeenRecentlyUnlocked} />}
```
C-gyorfi, line 118: Is it necessary to check !exceptionHasBeenRecentlyUnlocked since there is a separate lock for triggers/errors

agnieszkazbikowska-madetech: It's a result of some experimenting, in same cases I had the badge on 2 things.

PR #202, opened 2023-03-06
BICAWS-2712: 🐛  filter chips not removing on `cases allocated to me`

donnyhyon:
Bug fix to remove query params from by removing the `cases locked to me` filter chip from the filter panel. 

Co-authored-by: joe-fol <joe.folkard@madetech.com>

### DIFF
--- a/cypress/e2e/filter/filterChips.cy.ts
+++ b/cypress/e2e/filter/filterChips.cy.ts
@@ -392,9 +392,9 @@ describe("Case list", () => {
       })
     })
 
-    it("Should display the 'Locked to me' filter chip when selected", () => {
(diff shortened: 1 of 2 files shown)


src/features/CourtCaseFilters/AppliedFilters.tsx
```diff
@@ -21,7 +21,7 @@ interface Props {
     urgency?: string | null
     locked?: string | null
     caseState?: string | null
-    myCases?: boolean
+    myCases?: string | null
```
C-gyorfi, line 24: Can we pass this value as a boolean, since its a "true" or "false"?

monotypical (approved):
Nice work 🎉

C-gyorfi (approved):

PR #205, opened 2023-03-09
BICAWS-2675: Filter cases by SLA

C-gyorfi:
### Context
As a user, I would like to filter by case age(SLA) and know how many cases each day has so I can choose where the highest priority is.

### Changes
- Altered the named date range filtering section to let users filter by multiple days(SLA)
- Updated `listCourtCases` use case to allow filtering by multiple date ranges
- Added a new function(`getCountOfCasesByCaseAge`) to fetch the number of cases for each day. This is used for generating the labels for each case age option (see the screenshot)
- A few refactorings:
  - Rename `custom date range` to `date range` according to the design
  - Renaming date to caseAge in filter methods
  - Use `dateRange` instead of `from` and `to` params where possible
 - Deleted some skipped tests that were already fixed and tested in previous bug fixes

### Screenshot:
image


cypress/support/helpers.ts
```diff
@@ -17,3 +17,19 @@ export const confirmMultipleFieldsNotDisplayed = (fields: string[]) => {
     cy.contains(field).should("not.exist")
   })
 }
+
+export const removeFilterChip = () => {
```
monotypical, line 21: These helpers are really useful

src/types/CaseListQueryParams.ts
```diff
@@ -2,13 +2,18 @@ export type QueryOrder = "asc" | "desc" | undefined
 
 export type Reason = "Triggers" | "Exceptions" | "Bails"
 
-export type NamedCourtDateRanges = "Today" | undefined
+export type CaseAge = "Today" | "Yesterday" | "Day 2" | "Day 3"
```
monotypical, line 5: I think we need more case ages than these, the design has a scroll bar which would show more case ages. You can check with the designers for what case ages we want as options but I think we'll want at least 14 days

src/types/CountOfCasesByCaseAgeResult.ts
```diff
@@ -0,0 +1,6 @@
+export type CountOfCasesByCaseAgeResult = {
```
monotypical, line 1: Since we already have a `CaseAge` type, I wonder whether we could use a more generic type like `{[key: CaseAge]: number}` instead of duplicating all the case age options here

src/utils/caseAgeOptions.ts
```diff
@@ -0,0 +1,18 @@
...
+  },
+  Yesterday: () => {
+    return { from: subDays(new Date(), 1), to: subDays(new Date(), 1) }
+  },
+  "Day 2": () => {
+    return { from: subDays(new Date(), 2), to: subDays(new Date(), 2) }
+  },
+  "Day 3": () => {
+    return { from: subDays(new Date(), 3), to: subDays(new Date(), 3) }
+  }
```
monotypical, line 17: As mentioned in other comments, we'll need more options here for older cases

monotypical (approved):

PR #238, opened 2023-04-24
Bicaws 2755 triggers and exceptions sidebar

emadkaramad:
In this PR:
- Added JSON files containing the info about triggers and exceptions extracted from help page
- Added 2 functions to get the trigger info and exception info by code
- Added sidebar component for triggers and exceptions
- Added Triggers component
- Added Exceptions component
- Added ActionLink component. This component is generates a similar link to Link component but it doesn't accept `href`. `href` in Link `<a>` causes a jump to the top of the screen
- Added Checkbox component to customise the style. The `govuk-react` Checkbox component size doesn't match the checkbox in our design
- Added `shortTriggerCode` field and function to populate the value to the `Trigger` entity. This field is the short version of the trigger code (e.g. TRPR0001 -> PR01)
- Updated `createDummyCase` to use an AHO with one error when `errorReport` field contains exceptions.

image
image


src/utils/getTriggerInfo/triggers.json
C-gyorfi, line 1: Nice one extracting the info into JSON. Would it make sense to add this to the data repo? We have[ trigger definitions there](https://github.com/ministryofjustice/bichard7-next-data/blob/main/output-data/data/trigger-definitions.json) with slightly different descriptions:

emadkaramad: Good shout! Yeah, it makes sense to move both triggers and exceptions file to data repo. I'll do that in a separate PR.

src/components/ActionLink.tsx
```diff
@@ -0,0 +1,32 @@
...
+    border: "none",
+    color: "#1d70b8",
+    cursor: "pointer",
+    padding: "0",
+    textAlign: "left",
+    textDecoration: "underline"
+  }
+})
+
+const ActionLink: React.FC<Props> = ({ children, className, onClick }) => {
```
agnieszkazbikowska-madetech, line 22: You could do `const ActionLink = ({ children, className, onClick }: Props) => {` here, it might be easier to read? Tbh it's a preference thing.

src/features/CourtCaseDetails/CourtCaseDetails.tsx
```diff
@@ -22,6 +23,9 @@ interface Props {
   triggersVisible: boolean
 }
 
+const sideBarWidth = "364px"
```
agnieszkazbikowska-madetech, line 26: I like this! This way we don't have any magic numbers.

src/features/CourtCaseDetails/Sidebar/Exceptions.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const fieldName = String(path[path.length - 1])
+  const fieldNameWords = fieldName.match(/([A-Z]+[a-z]+)/g)
+  const formattedFieldName = fieldNameWords
+    ? fieldNameWords[0] + fieldNameWords.join(" ").slice(fieldNameWords[0].length).toLowerCase()
+    : fieldName
+
+  return `${formattedFieldName} (${location})`
+}
+
+const Exceptions: React.FC<Props> = ({ aho }) => {
```
agnieszkazbikowska-madetech, line 47: We can write it the other way (I've added a comment above) & avoid using `React.FC`.

src/features/CourtCaseDetails/Sidebar/TriggersAndExceptions.tsx
```diff
@@ -0,0 +1,55 @@
...
+const useStyles = createUseStyles({
+  pointer: {
+    cursor: "pointer"
+  },
+  sideBarMarginTop: {
+    marginTop: "-41px"
+  }
+})
+
+const TriggersAndExceptions: React.FC<Props> = ({ courtCase, aho }) => {
```
agnieszkazbikowska-madetech, line 23: We can write it the other way (I've added a comment above) & avoid using `React.FC`.

agnieszkazbikowska-madetech (changes requested):
Nice one! It's a big piece of work. I left a few comments but very minor.

src/features/CourtCaseDetails/Sidebar/Triggers.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }
+})
+
+const Triggers: React.FC<Props> = ({ courtCase }) => {
+  const classes = useStyles()
+  const [selectedTriggerIds, setSelectedTriggerIds] = useState<number[]>([])
+  const unresolvedTriggers = courtCase.triggers.filter((trigger) => !trigger.resolvedBy)
+
+  const setTriggerSelection = (event: ChangeEvent) => {
+    const checkBox = event.target as HTMLInputElement
```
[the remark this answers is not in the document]
emadkaramad: ChangeEvent is a generic type. We can pass the HTML element type to it:
`ChangeEvent<HTMLInputElement>`

agnieszkazbikowska-madetech (approved):

PR #240, opened 2023-04-25
BICAWS-2752 Allow to sort by number of notes

joe-fol:
This PR is to allow us to sort by the number of notes on the  `notes` column in the case list page. This is a copy of a branch that was previous worked on by other members and went stale after we managed to fix the issue with typeORM.


monotypical (approved):
Great work! 🎉

PR #242, opened 2023-04-26
Bicaws 2755 triggers and exceptions navigation link

emadkaramad:


src/types/NavigationLocations.ts
```diff
@@ -0,0 +1,7 @@
+type NavigationLocations =
```
monotypical, line 1: Since a value of this type represents a single location that is being viewed, I wonder if we want to use the singular name `NavigationLocation` here instead?

src/types/CaseDetailsTabs.ts
```diff
@@ -0,0 +1,3 @@
+type CaseDetailsTabs = "Defendant" | "Hearing" | "Case information" | "Offences" | "Notes" | "PNC errors"
```
monotypical, line 1: Similar here for plural/singular names

src/features/CourtCaseDetails/Sidebar/Exceptions.tsx
```diff
@@ -23,44 +27,33 @@ const useStyles = createUseStyles({
...
-    : fieldName
+const Exceptions = ({ aho, onNavigate }: Props) => {
+  const classes = useStyles()
 
-  if (location) {
-    location = ` (${location})`
+  const handleClick = (tab?: CaseDetailsTabs, offenceOrderIndex?: number) => {
+    switch (tab) {
+      case "Offences":
+        onNavigate({ location: "Case Details > Offences", args: { offenceOrderIndex } })
```
monotypical, line 36: I like this pattern for navigating between pages within sub-components! 😄

monotypical (approved):
Great work Emad! This feels really nice to use and I think is going to be something users love instead of having to work out what part of a case a trigger relates to! 🎉

monotypical (approved):

PR #243, opened 2023-04-26
Update visible cases query filter

C-gyorfi:
### Context
Currently, the list of court cases only filtered by `visible forces`, but instead `visible courts` or `visible forces` should filter it, [see the legacy codebase for reference.](https://github.com/ministryofjustice/bichard7-next/blob/main/bichard-backend/src/main/java/uk/gov/ocjr/mtu/br7/errorlistmanager/data/RecordSelectionSQLGenerator.java#L669)

### Changes
- Add a higher level `courtCasesByOrganisationUnitQuery` that will call both `courtCasesByVisibleForcesQuery` and `courtCasesByVisibleCourtsQuery`
- Use `courtCasesByOrganisationUnitQuery` where `courtCasesByVisibleForcesQuery` was used
- Update all tests(most of the change lies here since we used `courtCasesByVisibleForcesQuery` in most queries)


test/services/courtCasesByOrganisationUnitFilterQuery.integration.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  beforeAll(async () => {
+    dataSource = await getDataSource()
+  })
+
+  beforeEach(async () => {
+    await deleteFromEntity(CourtCase)
+    jest.resetAllMocks()
+    jest.clearAllMocks()
+    ;(courtCasesByVisibleCourtsQuery as jest.Mock).mockImplementation(
+      jest.requireActual("services/queries/courtCasesByVisibleCourtsQuery").default
```
emadkaramad, line 29: Nice way of restoring a module to its original implementation. 👍

emadkaramad (approved):

PR #244, opened 2023-04-26
2696 Update the look of the case list table

agnieszkazbikowska-madetech:
- Make background colours interchangeable case by case
- Add line separator between cases
- Add a grey line on the left
- Move tableHead to a new component - CourtCaseListTableHeader
- Reorganise the file structure to make it more logical


src/features/CourtCaseList/CourtCaseList.tsx
```diff
@@ -1,93 +1,38 @@
...
 
 interface Props {
   courtCases: CourtCase[]
   order?: QueryOrder
   currentUser: User
 }
 
-const CourtCaseList: React.FC<Props> = ({ courtCases, order = "asc", currentUser }: Props) => {
-  const { basePath, query } = useRouter()
+const CourtCaseList = ({ courtCases, order = "asc", currentUser }: Props) => {
```
monotypical, line 16: Do we want to make this function of type `React.FC<Props>` again? Doing so will tell us when we're passing something that doesn't match the expected argument types, and trying locally the compiler has no issue with the types

monotypical (approved):
Nice! The case list page looks very snazzy now 🙌

src/features/CourtCaseList/CourtCaseList.tsx
agnieszkazbikowska-madetech: I did some reading about it. I don't mind it either tbh. I think that TS still underlines wrong props in red if we leave it without `React.FC<Props>`.

PR #255, opened 2023-05-09
Ensure case visibility filter checks both visible force/court codes against org_for_police_filter and court_code

C-gyorfi:
### Context
Previously `courtCasesByOrganisationUnitQuery` checked the Users' `visibleCourts` codes against `courtCode` field and  `visibleForces` codes against the `orgPiliceFilter` field to determine whether a case should be visible to the user or not.
While this behaviour would make sense, the old application behaves slightly differently: it generates a combined `inclusion list` from visible courts and visible forces, then uses this combined list to find a matching code in `court_code` or `org_for_police_filter fields`

Changes:
- Update `courtCasesByOrganisationUnitQuery` to capture the described behaviour


cypress/e2e/filter/filterCases.cy.ts
```diff
@@ -43,7 +43,7 @@ describe("Case list", () => {
         users: [
           {
             username: "Bichard01",
-            visibleForces: ["011111"],
+            visibleForces: ["0011111"],
```
C-gyorfi, line 46: Added leading 0 because previously, the transformer function in the User entity trimmed the leading 0 from force codes when reading from the DB. With this PR we only trim the leading 0 from the codes when checking the case visibility [just like old Bichard](https://github.com/ministryofjustice/bichard7-next/blob/main/bichard-backend/src/main/java/uk/gov/ocjr/mtu/br7/errorlistmanager/data/RecordSelectionSQLGenerator.java#L704).

monotypical (approved):

PR #256, opened 2023-05-11
BICAWS-2856: Display all notes in a table on the Notes tab of the case details page

agnieszkazbikowska-madetech:
- create section in the court case details to show notes,
- add a cy test to check te notes tab,
- simplify cy tests related to notes.


cypress/e2e/index.cy.ts
```diff
@@ -11,7 +11,7 @@ import {
 } from "../support/helpers"
 import logAccessibilityViolations from "../support/logAccessibilityViolations"
 
-const loginAndGoToUrl = (emailAddress = "bichard01@example.com", url = "/bichard") => {
+export const loginAndGoToUrl = (emailAddress = "bichard01@example.com", url = "/bichard") => {
```
C-gyorfi, line 14: Should this go to the helpers?

agnieszkazbikowska-madetech: It should. I'll move it. Thanks!

PR #259, opened 2023-05-15
BICAWS-2220 Mark a case as manually resolved

C-gyorfi:
### Context
Implement the “Mark As Manually Resolved” functionality in the new UI. The focus of this ticket is functionality rather than a polished UX.

### Changes
- Capture the logic of [marking a case as manually resolved](https://github.com/ministryofjustice/bichard7-next/blob/main/bichard-backend/src/main/java/uk/gov/ocjr/mtu/br7/errorlistdialogue/actions/ConfirmMarkAsResolvedAction.java#L82)
- Add a placeholder resolve page and cypress tests(should be updated once we have a design)

### Next steps:
- Add audit log messages when a case is manually resolved(Old bichard pushes messages to GENERAL_EVENT_QUEUE). There is a ticket to [add audit logs to the new UI](https://dsdmoj.atlassian.net/browse/BICAWS-2777)
- Finalise the UI and update cypress tests

### DIFF
--- a/test/services/resolveCourtCase.integration.test.ts
+++ b/test/services/resolveCourtCase.integration.test.ts
@@ -81,7 +81,9 @@ describe("resolveCourtCase", () => {
         {
           errorLockedByUsername: resolverUsername,
           triggerLockedByUsername: resolverUsername,
-          orgForPoliceFilter: visibleForce
+          orgForPoliceFilter: visibleForce,
+          errorStatus: "Unresolved",
+          errorCount: 4
         }
       ])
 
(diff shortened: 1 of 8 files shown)


test/services/resolveCourtCase.integration.test.ts
```diff
@@ -0,0 +1,386 @@
...
+  beforeEach(async () => {
+    await deleteFromEntity(CourtCase)
+  })
+
+  afterAll(async () => {
+    await dataSource.destroy()
+  })
+
+  describe("When there aren't any unresolved triggers", () => {
+    it("should call cases by organisation unit query", async () => {
```
richardcane, line 65: Does this need to sit outside this parent block?

test/services/resolveCourtCase.integration.test.ts
```diff
@@ -0,0 +1,386 @@
...
+        .getMany()
+        .catch((error: Error) => error)
+      const courtCases = records as CourtCase[]
+
+      expect(courtCases).toHaveLength(1)
+      expect(courtCases[0].errorId).toEqual(caseId)
+      expect(courtCases[0].errorStatus).toEqual("Unresolved")
+    })
+
+    it("Should not resolve a case when the error is locked by another user", async () => {
```
richardcane, line 209:
```suggestion
    it("Should not resolve a case when the case is locked by another user", async () => {
```

test/services/resolveCourtCase.integration.test.ts
```diff
@@ -0,0 +1,386 @@
...
+      }
+      expect(result).toEqual(Error(`Failed to update record with some error`))
+
+      const record = await dataSource.getRepository(CourtCase).findOne({ where: { errorId: 0 } })
+      const actualCourtCase = record as CourtCase
+
+      expectToBeUnresolved(actualCourtCase)
+    })
+  })
+})
```
richardcane, line 386: Should there be a block that describes the situation in which there are no exceptions on a case, only triggers?

agnieszkazbikowska-madetech (approved):

PR #260, opened 2023-05-16
Install and configure eslint-plugin-no-only-tests

richardcane:
image


agnieszkazbikowska-madetech (approved):

richardcane:
Noticed that the lint step in the build wasn't actually applying linting to any of the code, so am currently trying to fix that.

PR #271, opened 2023-06-02
BICAWS-2872: Reallocate case page

emadkaramad:
This PR updates the reallocate case page to match the design.


src/features/CourtCaseDetails/Sidebar/Trigger.tsx
richardcane, line 1: I've done a bigger overhaul for this group of components in #272

src/utils/reallocation-forces.json
monotypical, line 1: This list could fall out of sync with what's in the data repo when we add or remove forces from that. It might be better to use that list, and filter out any forces that we shouldn't assign cases to (if there are any)

src/features/CourtCaseDetails/Sidebar/Trigger.tsx
emadkaramad: cool - I will revert the file changes then to avoid conflict with your PR.

src/utils/reallocation-forces.json
emadkaramad: Good shout Alice. That's a good point. I put them in a separate file similar to the old Bichard as some of the force names are shortened and not exactly the same as the names in standing data. But I think that should be fine to use the force names in standing data.
I'll update that to just keep the force codes for reallocation in this repo and read the names from the standing data.

PR #277, opened 2023-06-06
Bicaws 2858 add user notes to notes tab

joe-fol:
This PR allows a user to add a note from the `Notes` tab. Previously it routed to its on `Notes` page but now it stays in the Notes view:

- Update the current component `addNoteForm` to reflect design
- Update test to use new behavior. e.g cant add a note when locked to someone else
- Added helper function for testing and moved some functions to keep thing dry
- Changed behavior so when you submit a form, on load it stays on the notes tab  
- Relocated some files as we no longer need to be on the notes page
- Moved the constant variable `MAX_NOTE_LENGTH` to a config file


cypress/e2e/court-cases/CourtCaseDetails.Notes.cy.ts
```diff
@@ -158,6 +182,114 @@ describe("Court case details", () => {
...
+  it("should not display the notes text area if the case is locked to another users", () => {
+    cy.task("insertCourtCasesWithNotesAndLock", {
+      caseNotes: [[]],
+      force: "02"
+    })
+    cy.login("bichard02@example.com", "password")
+    cy.visit("/bichard/court-cases/0")
+
+    clickTab("Notes")
+    cy.contains("Case has no notes.")
```
monotypical, line 195: This test doesn't seem to actually test that the notes text area isn't present. It might be worth using `cy.get(<notes text area selector>).should("not.exist")` to check this

src/features/CourtCaseDetails/Tabs/Panels/Notes/AddNoteForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+            <TextArea
+              input={{
+                id: "note-text",
+                name: "noteText",
+                rows: 5,
+                maxLength: MAX_NOTE_LENGTH,
+                onInput: handleOnNoteChange
+              }}
+              meta={{
+                error: "Input message into text box",
```
monotypical, line 45: This error message might be clearer to users if we're more direct about what the problem is, e.g. "The note cannot be empty"

cypress/e2e/court-cases/CourtCaseDetails.Notes.cy.ts
```diff
@@ -158,6 +182,114 @@ describe("Court case details", () => {
...
+    cy.get("H1").should("have.text", "Case details")
+    clickTab("Notes")
+    cy.findByText("Case has no notes.").should("exist")
+  })
+
+  it("should return 404 for a case that this user can not see", () => {
+    cy.task("insertCourtCasesWithFields", [{ orgForPoliceFilter: "02" }])
+    cy.login("bichard01@example.com", "password")
+
+    cy.request({
```
monotypical, line 275: I don't think these tests are relevant anymore since the notes are in a tab on the case page and not a separate page. We already test this on the case details page at https://github.com/ministryofjustice/bichard7-next-ui/blob/8b53cea5825af957ccafe65596bb55befffb05cb/cypress/e2e/court-cases/CourtCaseDetails.cy.ts#L184

src/features/CourtCaseDetails/CourtCaseDetails.tsx
```diff
@@ -36,6 +37,20 @@ const CourtCaseDetails: React.FC<Props> = ({ courtCase, aho, lockedByAnotherUser
   const [selectedOffenceIndex, setSelectedOffenceIndex] = useState<number | undefined>(undefined)
   const classes = useStyles()
 
+  useEffect(() => {
```
monotypical, line 40: this is a really nice usability improvement 🎉

monotypical (commented):
I've left a couple of comments, otherwise LGTM

monotypical (approved):

PR #278, opened 2023-06-06
BICAWS-2539: Persist applied filters in case list

emadkaramad:


src/pages/index.tsx
```diff
@@ -100,6 +107,12 @@ export const getServerSideProps = withMultipleServerSideProps(
           throw lockResult
         }
       }
+    } else if (!isPost(req) && req.url) {
```
monotypical, line 110: I'm not sure we want to ignore post requests for this, on the case list page a supervisor could:
1) apply some filter
2) unlock triggers on a case, sending a post request to unlock the page and receiving the case list page back with the triggers for that case unlocked

In this scenario the filters would not be persisted and the supervisor would see the unfiltered case list page

monotypical (commented):
Looks great, I've left a comment about an edge case to consider

src/pages/index.tsx
emadkaramad: I just checked it and it works fine. It's because the unlock trigger action posts back to the same URL which has all fields in the query string.
But it is a good idea to not limit it to only GET method, so if we changed the form action to post back to the URL without query string, it redirects the user back to the URL in the cookie.

monotypical (approved):

PR #287, opened 2023-06-13
Added 401 cypress test for authjwt

donnyhyon:


donnyhyon:
This logic is handed by nginx and is an end point on the user-service. 

This logic has been applied into the user service repo.

Closing branch as no longer needed.

PR #293, opened 2023-06-15
BICAWS-2912: Lock case when defendant name is clicked

ian-antking:


src/features/CourtCaseList/CourtCaseListEntry/CaseDetailsRow/CaseDetailsRow.tsx
```diff
@@ -96,7 +110,11 @@ export const CaseDetailsRow = ({
           </ConditionalRender>
         </Table.Cell>
         <Table.Cell className={caseDetailsCellClass}>
-          <Link href={caseDetailsPath(errorId)} id={`Case details for ${defendantName}`}>
+          <Link
+            onClick={async (event) => lockAndOrViewCase(event)}
+            href="" // emply href enables expected mouseover behaviour
```
emadkaramad, line 115: I wonder if we can use `<ActionLink>` component here. It does generate a link-like button without the need to set the href attribute.

emadkaramad (dismissed):

src/features/CourtCaseList/CourtCaseListEntry/CaseDetailsRow/CaseDetailsRow.tsx
ian-antking: :eyes:

ian-antking: Have swapped it out. Basically the same implementation, but without the need for an empty href and preventing default behaviour.

emadkaramad (approved):

richardcane (approved):

ian-antking:
Reverted from using an ActionLink because it changed the element type from `a` to `button`, thus breaking the tests. Using a button for navigation also affects accessibility.

PR #296, opened 2023-06-19
Role based Reason filtering

richardcane:


cypress/e2e/filter/reasons.roles.cy.ts
```diff
@@ -0,0 +1,94 @@
...
+        email: email,
+        password: hashedPassword
+      }
+    ],
+    userGroups: [Groups.NewUI, ...groups]
+  })
+  cy.login(email, "password")
+}
+
+function navigateAndShowFilters() {
```
C-gyorfi, line 23: I'm going to be pedantic here:
Would we prefer using arrow functions over "traditional functions"? I think we using arrow functions most of times in this codebase. While in some cases, it just a style choice. For instance I like arrow functions because they have a more concise syntax and quite readable. But there are technical differences too:

- const prevents reassignment but function not
- arrow functions has to be declared before calling it, while functions can be defined after calling them

src/types/CaseListQueryParams.ts
```diff
@@ -1,6 +1,11 @@
 export type QueryOrder = "asc" | "desc"
 
-export type Reason = "Triggers" | "Exceptions" | "Bails"
+export type Reason = Reasons
+export enum Reasons {
```
C-gyorfi, line 4: I wonder if we could avoid having both Reason and Reasons defined, I get that one is a type the other is an enum, but it was quite unclear from the naming until I looked it up

C-gyorfi (approved):
Looks good!

src/types/CaseListQueryParams.ts
richardcane: Yeah I don't like this either.

C-gyorfi: Nice one! 🎉

C-gyorfi (approved):
:shipit:

PR #299, opened 2023-06-20
NextJS standalone build configuration

emadkaramad:
Implements the [standalone build configuration for nextjs](https://nextjs.org/docs/app/api-reference/next-config-js/output)

CircleCI builds have been becoming less and less performant over time. We looked at trying to parallelise jobs in the workflow, but the e2e tests running against a Docker image meant we had to choose between:

- Caching
  - CircleCI DLC is flaky and difficult to predict.
- Registry
  - This is still an option but it requires us to think about strategy for generating lots of preproduction images.
- Artifacts
  - Given the large size of the generated images, we determined this would be infeasible unless we could reduce the image size.

In order to reduce the image size, we looked at the size of node modules being imported and discovered that much of the image size was accounted for by standing data packages being installed as dependencies of the `bichard7-next-core` package.

Rather than delete those dependencies directly, we opted to attempt to use the standalone output configuration of nextjs to generate the smallest possible production-ready nextjs application. This resulted in a 10x reduction in image size, which makes the passing of the image as an artifact from job to job in the CircleCI workflow feasible.


Dockerfile
```diff
@@ -40,11 +40,15 @@ ENV NODE_ENV=production
...
-COPY --from=app_builder /src/ui/node_modules ./node_modules
-COPY --from=app_builder /src/ui/package*.json ./
 COPY --from=app_builder /src/ui/next.config.js ./
+COPY --from=app_builder /src/ui/package*.json ./
 COPY --from=app_builder /src/ui/public ./public
-COPY --from=app_builder --chown=nextjs:nodejs /src/ui/.next ./.next
+COPY --from=app_builder /src/ui/.next/standalone .
+COPY --from=app_builder /src/ui/.next/static ./.next/static
+# Copy standing data package to Core repository as core dependency
+COPY --from=app_builder \
```
richardcane, line 49: Just to leave a more detailed comment - the NextJS standalone build analyses imports and it determines that these deps aren't required and deletes them.

https://github.com/ministryofjustice/bichard7-next-core/blob/13c008fd3d919c705fb851ce02716854d7f77074/src/lib/requireStandingData.ts

C-gyorfi (approved):

PR #304, opened 2023-06-23
Add "Leave and lock", "Leave and unlock" buttons

monotypical:
This PR:
* Adds "Leave and lock", "Leave and unlock" buttons when a user has the triggers locked
* Adds a "Return to case list" button when the user does not have a case locked
* Moves the reallocation button with the other buttons in the header of the court case details page
* Extracts components above the summary box into a `Heading` component on the case details page
* Adds cypress tests for buttons in their own file
* Prevents unlockTrigger and unlockException params from being saved in the cookie
* Fixes spacing around the case details header to match design
* Adds util functions for checking whether a case is locked after the case/user entities have been serialised

Screenshots:
Locked by the current user:
Screenshot 2023-06-23 at 17 35 15

Locked 
Screenshot 2023-06-23 at 17 34 27
by another user:


cypress/e2e/court-cases/court-case-details/CourtCaseDetails.cy.ts
```diff
@@ -270,74 +265,6 @@ describe("Court case details", () => {
     cy.url().should("match", /\/users/)
   })
 
-  it("should lock a case when a user views a case details page", () => {
```
C-gyorfi, line 273: Isn't this the expected behaviour anymore?

C-gyorfi (dismissed):

cypress/e2e/court-cases/court-case-details/CourtCaseDetails.cy.ts
C-gyorfi: Awesome! 🚀

PR #306, opened 2023-06-26
Add audit log events when unlocking a case

C-gyorfi:
### Changes:
- [Introduce a higher level unlockCourtCase function](https://github.com/ministryofjustice/bichard7-next-ui/pull/306/commits/a462c57ff830e1d25bbf102ead819c7212225dc7):
  - This will be responsible for running the transaction and calling `updateLockStatusToUnlocked` and `storeAuditLogEvents`
- Renamed the old `unlockCourtCase` function to `updateLockStatusToUnlocked`, this is a lower-level operation that is called in other transactions(e.g resolveCourtCase or reallocateCourtCase)


src/services/reallocateCourtCaseToForce.ts
```diff
@@ -60,7 +61,13 @@ const reallocateCourtCaseToForce = async (
         }
       }
 
-      const unlockResult = await unlockCourtCase(entityManager, +courtCaseId, user)
+      const unlockResult = await updateLockStatusToUnlocked(
+        entityManager,
+        +courtCaseId,
+        user,
+        UnlockReason.TriggerAndException,
+        [] //TODO pass an actual audit log events array
```
monotypical, line 69: Does this need to be completed?

src/services/resubmitCourtCase/resubmitCourtCase.ts
```diff
@@ -53,7 +53,13 @@ const resubmitCourtCase = async (
           return statusResult
         }
 
-        const unlockResult = await unlockCourtCase(entityManager, +courtCaseId, currentUser)
+        const unlockResult = await updateLockStatusToUnlocked(
+          entityManager,
+          +courtCaseId,
+          currentUser,
+          UnlockReason.TriggerAndException,
+          [] //TODO pass an actual audit log events array
```
monotypical, line 61: Same here, does this need to be completed for this PR?

src/services/reallocateCourtCaseToForce.ts
C-gyorfi: We decided to do a PR just for unlocking because it involved quite a bit of a refactor, more PR coming soon!

monotypical (approved):

PR #314, opened 2023-06-29
Update locks on case details page

monotypical:
This PR updates the locks on the case details page to match the design, and adds a "view only" badge when no locks are held

Screenshot 2023-06-29 at 11 04 45


cypress/e2e/court-cases/court-case-details/CourtCaseDetails.cy.ts
```diff
@@ -76,7 +76,6 @@ describe("Court case details", () => {
 
     cy.contains("Case00000")
     cy.contains("Magistrates' Courts Essex Basildon")
-    cy.contains("Urgent")
```
monotypical, line 79: Moved this test to `cypress/e2e/court-cases/court-case-details/Header.cy.ts`

cypress/e2e/court-cases/court-case-list/CourtCaseList.cy.ts
```diff
@@ -41,9 +41,12 @@ describe("Court case details", () => {
     cy.visit("/bichard")
 
     cy.contains("a", "NAME Defendant").click()
-    cy.findByText("Case locked by another user").should("not.exist")
-    cy.contains("p", "Trigger locked by: Bichard02").should("exist")
-    cy.contains("p", "Error locked by: Bichard02").should("exist")
+
+    cy.get(".view-only-badge").should("not.exist")
```
monotypical, line 45: Updated these tests to use new format of locks on case details page

src/components/Badge.tsx
```diff
@@ -0,0 +1,18 @@
+import ConditionalRender from "components/ConditionalRender"
+
+interface BadgeProps {
+  isRendered: boolean
+  className?: string
+  colour: string
+  label: string
+}
+
+const Badge = ({ isRendered, className, colour, label }: BadgeProps) => {
```
richardcane, line 10: This is probably quite a nice pattern in general for things that need showing and hiding based on roles etc

richardcane (approved):
All good!

src/services/entities/User.ts
```diff
@@ -39,14 +39,20 @@ export default class User extends BaseEntity {
...
+        group === UserGroup.Allocator ||
+        group === UserGroup.Supervisor
     )
   }
 
   get canLockExceptions() {
     return this.groups.some(
       (group) =>
-        group === "ExceptionHandler" || group === "GeneralHandler" || group === "Allocator" || group === "Supervisor"
+        group === UserGroup.ExceptionHandler ||
```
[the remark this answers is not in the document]
monotypical: good catch, I've replaced the functions on the user entity with the ones from `utils/userPermissions` now. I tried to get the user entity to call the other ones but that causes an import cycle as the utils file depends on `User`

richardcane (approved):

PR #318, opened 2023-06-30
Bicaws 2777 trigger resolved event

emadkaramad:


src/services/unlockCourtCase.ts
```diff
@@ -13,11 +12,15 @@ const unlockCourtCase = async (
   courtCaseId: number,
   user: User,
   unlockReason: UnlockReason
-): Promise<UpdateResult | Error> => {
+): Promise<UpdateResult | Error | undefined> => {
```
C-gyorfi, line 15: This is not pretty, but we are going to refactor the return types of all the service functions in the next PR

src/services/updateLockStatusToUnlocked.ts
```diff
@@ -8,14 +8,15 @@ import getAuditLogEvent from "@moj-bichard7-developers/bichard7-next-core/build/
...
+import getCourtCase from "./getCourtCase"
 
 const updateLockStatusToUnlocked = async (
   dataSource: EntityManager,
   courtCaseId: number,
   user: User,
   unlockReason: UnlockReason,
   events: AuditLogEvent[]
-): Promise<UpdateResult | Error> => {
+): Promise<UpdateResult | Error | undefined> => {
```
C-gyorfi, line 19: Same here, refactor PR is coming

src/services/reallocateCourtCaseToForce.ts
```diff
@@ -8,80 +8,110 @@ import insertNotes from "./insertNotes"
...
-            userId: user.username
-          }
-        ])
-
-        if (isError(addUserNoteResult)) {
-          throw addUserNoteResult
-        }
-      }
+    events.push(
+      getAuditLogEvent("information", "Hearing outcome reallocated by user", "Bichard New UI", {
```
monotypical, line 79: suggestion: I think it would be good to have the event category, event type and event source be defined in a type or enum somewhere, which would help prevent us from making typos in event codes and let us update the codes/source easily. I don't know if this is something that's on your radar for the refactor PR you've mentioned

src/services/reallocateCourtCaseToForce.ts
```diff
@@ -8,80 +8,110 @@ import insertNotes from "./insertNotes"
...
+    const events: AuditLogEvent[] = []
+
+    const courtCase = await getCourtCase(entityManager, courtCaseId)
+
+    if (isError(courtCase)) {
+      throw courtCase
+    }
+
+    if (!courtCase) {
+      throw new Error("Failed to unlock: Case not found")
```
monotypical, line 36: I think this error message isn't accurate since we're failing to reallocate the case here, not unlock it

monotypical (approved):
Looks good!

src/services/reallocateCourtCaseToForce.ts
emadkaramad: Thanks Alice! Yes, this is something that we thought about and we think it should exist in the core repository so we can then use it in the UI repo. We will open separate PRs for these improvements.

PR #326, opened 2023-07-05
Refactor existing user permissions checks

monotypical:
These now all use properties on the `User` entity. These are included in the serialised object passed in props by overriding `toJSON` of the base entity and including all properties with getters


src/services/entities/BaseEntity.ts
```diff
@@ -1,5 +1,28 @@
...
   serialize() {
     return JSON.parse(JSON.stringify(this))
   }
+
+  // Add properties with getters to the JSON object
+  toJSON() {
+    const proto = Object.getPrototypeOf(this)
+    const jsonObj: any = Object.assign({}, this)
+
+    Object.entries(Object.getOwnPropertyDescriptors(proto))
```
C-gyorfi, line 13: 👍

src/services/entities/User.ts
```diff
@@ -35,4 +35,34 @@ export default class User extends BaseEntity {
   featureFlags!: KeyValuePair<string, boolean>
 
   groups: UserGroup[] = []
+
+  get hasAccessToTriggers() {
```
C-gyorfi, line 39: So much cleaner 👏

C-gyorfi (approved):

cypress/e2e/court-cases/court-case-details/TriggersAndExceptions.cy.ts
```diff
@@ -378,7 +378,7 @@ describe("Triggers and exceptions tabs", () => {
   })
 
   it("should only show triggers to Trigger Handlers", () => {
-    newUserLogin({ groups: [UserGroup.TriggerHandler] })
+    newUserLogin({ groups: [UserGroup.TriggerHandler], hasAccessToTriggers: true })
```
richardcane, line 381: I'm happy to approve this but I wonder if it's something that maybe we'd want to infer from `newUserLogin` from the groups passed in 🤔

richardcane: I had a niggling feeling when we removed the shared implementations of those functions and this might be why

richardcane (dismissed):

cypress/e2e/court-cases/court-case-details/TriggersAndExceptions.cy.ts
monotypical: that's a really good point, I'll update put the util functions back in and use those

monotypical: Changed in https://github.com/ministryofjustice/bichard7-next-ui/pull/326/commits/b1c9a5803e80c5b5f1953114b1084e38424f6a9e

richardcane (approved):

PR #331, opened 2023-07-07
Bicaws-2941: Fix always truthy assertion and return exception messages from service

ian-antking:
This `if` statement would never trigger, as objects are truthy, even when empty.

```typescript
        if (!addNoteResult) {
          return notSuccessful("Case is locked by another user")
        }
```

This commit changes the condition to check that the `isSuccessful` property is not false. If it is false, then it will return the `validationError` or the `Exception.message`. If neither is truthy then it will default to an empty string.


monotypical (approved):

monotypical:
Good catch!

PR #339, opened 2023-07-13
Unlock triggers when all triggers are resolved

C-gyorfi:


test/services/resolveTriggers.integration.test.ts
```diff
@@ -50,6 +50,18 @@ describe("resolveTriggers", () => {
   const createAllTriggersResolvedEvent = (triggers: string[], username = "triggerResolver01") =>
     createTriggersEvent("triggers.all-resolved", "All triggers marked as resolved", triggers, username)
 
+  const triggerUnlockedEvent = {
```
monotypical, line 53: I wonder if it's worth distinguishing why a trigger was unlocked here, i.e. in this case triggers were unlocked because they were all resolved, vs. a user unlocking them when returning to the case list

monotypical (dismissed):

test/services/resolveTriggers.integration.test.ts
C-gyorfi: I think it's a valid point and could be a good future improvement. I just double-checked, the legacy application doesn't differentiate the two unlock events, but that doesn't meant the new UI shouldn't!

monotypical (approved):

PR #351, opened 2023-08-01
Add permission flags to users to control what features are shown

monotypical:
* Replaced `hasAccessToTriggers`, `hasAccessToExceptions` and `isSupervisor` with a more general `hasAccessTo` mapping using the newly added `Feature`
* Corrected e2e test, users with no groups should see neither exceptions nor triggers
* Added a `UnlockOtherUsersCases` feature permission, only supervisors can do this


src/pages/index.tsx
```diff
@@ -117,7 +118,9 @@ export const getServerSideProps = withMultipleServerSideProps(
     }
 
     const resolvedByUsername =
-      validatedCaseState === "Resolved" && !currentUser.isSupervisor ? currentUser.username : undefined
+      validatedCaseState === "Resolved" && !currentUser.hasAccessTo[Feature.UnlockOtherUsersCases]
```
C-gyorfi, line 121: I think it was more readable with the `!currentUser.isSupervisor` here since we use the `resolvedByUsername` var to filter resolved cases visible to the user

C-gyorfi: Maybe we could define another feature something like `Feature.ListAllResolvedCases` to make it more explicit

test/helpers/hasAccessTo.ts
```diff
@@ -0,0 +1,8 @@
+import { UserGroup } from "types/UserGroup"
+import { userAccess } from "utils/userPermissions"
+
+export const hasAccessToAll = userAccess({ groups: Object.values(UserGroup) })
+export const hasAccessToNone = userAccess({ groups: [] })
```
C-gyorfi, line 5: I don't think we use this anywhere

test/services/listCourtCases.integration.test.ts
```diff
@@ -1570,43 +1568,37 @@ describe("listCourtCases", () => {
...
-      hasAccessToTriggers: false
+      hasAccessTo: userAccess({ groups: [] })
     } as Partial<User> as User
     const triggerHandlerUser = {
       visibleForces: [orgCode],
       visibleCourts: [],
       groups: [UserGroup.TriggerHandler],
-      hasAccessToExceptions: false,
-      hasAccessToTriggers: true
+      hasAccessTo: userAccess({ groups: [UserGroup.TriggerHandler] })
```
C-gyorfi, line 1577: Could possibly refactor these tests to make use of the defined helpers, for instance:
`hasAccessTo: generalHandlerHasAccessTo`

C-gyorfi (dismissed):

test/services/listCourtCases.integration.test.ts
monotypical: I've switched this over to use the helpers

test/helpers/hasAccessTo.ts
monotypical: I've used these in the `listCorutCases` integration test now

C-gyorfi (approved):

PR #357, opened 2023-08-07
fixed generated cases with nothing to do

Senny10:
insure that exceptions are always generated when there are only resolved triggers on a generated case.


monotypical (approved):
LGTM 🎉

PR #358, opened 2023-08-07
Fix defendant name links

emadkaramad:
This PR is to:
- update the defendant name link in case list page to set the `href` correctly
- remove the defendant name link ID attribute. The attribute was not needed and the value format was invalid.
- remove the logic to lock the case on the client-side and move it to the case details page. Now when user accesses the case details page, system tries to lock the exceptions and triggers. The `lock` parameter in the query string can only be used to unlock the case via a POST request now.

### DIFF
--- a/src/pages/court-cases/[courtCaseId]/index.tsx
+++ b/src/pages/court-cases/[courtCaseId]/index.tsx
@@ -38,12 +38,10 @@ export const getServerSideProps = withMultipleServerSideProps(
 
     let lockResult: UpdateResult | Error | undefined
 
-    if (isPost(req) && !!lock) {
-      if (lock === "false") {
-        lockResult = await unlockCourtCase(dataSource, +courtCaseId, currentUser, UnlockReason.TriggerAndException)
-      } else {
-        lockResult = await lockCourtCase(dataSource, +courtCaseId, currentUser)
-      }
+    if (isPost(req) && lock === "false") {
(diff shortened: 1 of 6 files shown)


src/pages/court-cases/[courtCaseId]/index.tsx
```diff
@@ -38,16 +39,26 @@ export const getServerSideProps = withMultipleServerSideProps(
     }
     const dataSource = await getDataSource()
 
-    let lockResult: UpdateResult | Error | undefined
+    let courtCase = await getCourtCaseByOrganisationUnit(dataSource, +courtCaseId, currentUser)
```
Comment (emadkaramad) [inline src/pages/court-cases/[courtCaseId]/index.tsx:42]: Pre-update fetch from the database to ensure that the court case exists and user has access to the case.

src/pages/court-cases/[courtCaseId]/index.tsx
```diff
@@ -106,7 +117,8 @@ export const getServerSideProps = withMultipleServerSideProps(
       }
     }
 
-    const courtCase = await getCourtCaseByOrganisationUnit(dataSource, +courtCaseId, currentUser)
+    // Fetch the record from the database after updates
+    courtCase = await getCourtCaseByOrganisationUnit(dataSource, +courtCaseId, currentUser)
```
Comment (emadkaramad) [inline src/pages/court-cases/[courtCaseId]/index.tsx:121]: Post-update fetch the court case from the database to ensure that we use the updated court case value to render the UI.

richrace (dismissed):
👍

C-gyorfi (dismissed):

richardcane (approved):

PR #360, opened 2023-08-09
Show different reason columns to different user roles

monotypical:
* Trigger handlers only see triggers in the case list, and exception handlers only see exceptions
* Cases with only triggers now display as one row, rather than having an empty reason column on the top row, and the trigger reason on a second row
* We now pass in the reason column and lock tag to each row on the case list page, and calculate what to display from `CourtCaseListEntry`
* Moved exceptions related columns to separate components in `ExceptionColumns` and the same for triggers

General handler view of a set of cases with mixed triggers and exceptions:
Screenshot 2023-08-09 at 11 05 38

Trigger handler view of the same cases:
Screenshot 2023-08-09 at 11 03 44

Exception handler view of the same cases:
Screenshot 2023-08-09 at 11 06 16


src/features/CourtCaseList/CourtCaseListEntry/ExceptionsColumns.tsx
```diff
@@ -0,0 +1,39 @@
...
+  exceptionsHaveBeenRecentlyUnlocked: boolean
+}) => {
+  const lockTag = canUnlockCase ? (
+    <LockedByTag lockedBy={errorLockedByFullName} unlockPath={unlockPath} />
+  ) : (
+    <LockedByTag lockedBy={errorLockedByFullName} />
+  )
+  return (
+    <>
+      {lockTag}
```
emadkaramad, line 35: I wonder if we can simplify `lockTag` to:
`<LockedByTag lockedBy={errorLockedByFullName} unlockPath={canUnlockCase ? unlockPath : undefined} />`

src/features/CourtCaseList/CourtCaseListEntry/TriggersColumns.tsx
```diff
@@ -0,0 +1,66 @@
...
+}: {
+  triggersLockedByUsername: string | null | undefined
+  triggersLockedByFullName: string | null | undefined
+  triggersHaveBeenRecentlyUnlocked: boolean
+  canUnlockCase: boolean
+  unlockPath: string
+}) => {
+  return (
+    <>
+      {canUnlockCase ? (
```
emadkaramad, line 57: Same here if we can simplify this check.

emadkaramad (approved):

src/features/CourtCaseList/CourtCaseListEntry/ExceptionsColumns.tsx
monotypical: Thanks emad, I've made that refactor now 🎉

src/features/CourtCaseList/CourtCaseListEntry/TriggersColumns.tsx
monotypical: Thanks emad, I've made that refactor now 🎉

emadkaramad (approved):

PR #365, opened 2023-08-16
Use Core to run dependencies in CI

richardcane:
- Brings CI config in line with config style established in tests repo
- Upgrades node to 16.20.2
- Removes build steps


.circleci/config.yml
```diff
@@ -1,206 +1,133 @@
...
+      - fetch_images:
+          IMAGES: nodejs
+      - build_message_forwarder
+      - build_audit_logging_images
+      - run:
+          name: Start bichard7 legacy
+          working_directory: ~/bichard7-next-core
+          command: npm run all-legacy
+    
+  start_specific_core_services:
```
C-gyorfi, line 73: I'll be pedantic here: Why core services? :D

C-gyorfi (dismissed):
🚀

.circleci/config.yml
richardcane: Fair point

C-gyorfi (approved):

PR #375, opened 2023-08-31
Bicaws 3037 add general feedback page

Senny10:
Created a general feedback page. This PR includes the following:

- Updated `cypress.config` to include 2 new functions `getAllFeedbacksFromDatabase()` and `clearAllFeedbacksFromDatabase()` found in `test/utils/manageFeedbackSurveys.ts`
- implemented general feedback page `feedback.tsx` and tested using Cypress test

image


src/pages/feedback.tsx
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    return { props }
+  }
+)
+
+interface Props {
+  user: User
+  previousPath: string
+  displayAnonymousMissingError: boolean
```
C-gyorfi, line 91: We could group these error attributes just like `selectedFields` or even better could just add a fields property with two attributes, for instance:
```
fields: {
   isAnonymous: {
      hasError: boolean,
      value: string
   },
   experience: {
      hasError: boolean,
      value: string
   }
   ...
}
```

src/pages/feedback.tsx
```diff
@@ -0,0 +1,232 @@
...
+  displayAnonymousMissingError,
+  displayExperienceMissingError,
+  displayfeedbackMissingError,
+  selectedFields
+}: Props) => {
+  const maxFeedbackNoteLength: number = 2000
+  const [noteRemainingLength, setNoteRemainingLength] = useState(maxFeedbackNoteLength)
+
+  const handleOnNoteChange: FormEventHandler<HTMLTextAreaElement> = (event) => {
+    setNoteRemainingLength(maxFeedbackNoteLength - event.currentTarget.value.length)
```
C-gyorfi, line 109: I believe the term "note" is coming from the note page, maybe we could use a term that is better describes the current context: setRemainingFeedbackLength, handleFeedbackChange, etc.

src/pages/feedback.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+const FeedbackPage: NextPage<Props> = ({
+  user,
+  previousPath,
+  displayAnonymousMissingError,
+  displayExperienceMissingError,
+  displayfeedbackMissingError,
+  selectedFields
+}: Props) => {
+  const maxFeedbackNoteLength: number = 2000
```
C-gyorfi, line 105: We could set the max feedback length in the config just as we do with note length

PR #393, opened 2023-09-21
Continue workflow in conductor when case is manually resolved

richardcane:
Call Conductor API to continue the `bichard_process` workflow when manually resolving a case. This only applies if `LEGACY_PHASE1` variable is `false`, i.e. we are using Conductor/core.


cypress.config.ts
```diff
@@ -135,5 +135,5 @@ export default defineConfig({
     }
   },
 
-  retries: 2
+  retries: process.env.CI ? 2 : 0
```
ammaar-limbada-madetech, line 138: This might need bringing to the attention of other devs so adding a comment - will stop Cypress from retrying locally and only retry when running in CI.

test/services/resolveCourtCase.integration.test.ts
```diff
@@ -532,6 +533,24 @@ describe("resolveCourtCase", () => {
...
+    it("Should return the error if fails to store audit logs", async () => {
+      const error = `Error while calling audit log API`
+      ;(storeAuditLogEvents as jest.Mock).mockImplementationOnce(() => new Error(error))
+
+      let result
+      try {
+        result = await resolveCourtCase(dataSource, courtCases[0], resolution, user)
+      } catch (e) {
+        result = e as Error
+      }
```
emadkaramad, line 545: Nitpick: It could be simplified as `const result = await resolveCourtCase(dataSource, courtCases[0], resolution, user).catch((error) => error as Error)`

cypress.config.ts
emadkaramad: Nice! 🎉

emadkaramad (approved):

emadkaramad (approved):

PR #398, opened 2023-09-25
Column heading icon's are now clickable

joe-fol:
This P.R makes the icon in the column headings clickable, it was picked up in a round of user testing. 
I have updated some cypress tests to have a variety of interactions e.g click column title or icon

[Kooha-2023-09-26-09-24-26.webm](https://github.com/ministryofjustice/bichard7-next-ui/assets/65510707/7c0c2db6-c792-4684-95ba-e324daa28ce6)


cypress/e2e/court-cases-index/sortingCases.cy.ts
```diff
@@ -115,7 +113,7 @@ describe("Sorting cases", () => {
     checkCasesOrder([1, 3, 2, 0])
 
     // Sort descending by court name
-    cy.get("#court-name-sort").click()
+    cy.get("#court-name-sort").find(".upArrow").click()
```
richrace, line 116: I would add extra tests for the arrow and the `#court-name-sort` element. They test two different behaviours. Does that make sense?

And I test the `.downArrow` as well.

richrace (approved):

PR #404, opened 2023-10-03
reports button is only visible for Supervisors

BushraAbdullahi:
This pr is for making sure the Reports link is only visible for supervisors


monotypical (approved):

src/components/NavBar/NavBar.tsx
```diff
@@ -36,14 +36,15 @@ const UserManagementNavItem: React.FC<NavBarProps> = (groups) => {
 
 const NavBar: React.FC<NavBarProps> = ({ groups }) => {
   const classes = useCustomStyles()
+  const isSupervisor = groups.some((group) => group === UserGroup.Supervisor)
```
C-gyorfi, line 39: Could we extend `hasAccess` function on the user entity to determine this permission? There is an `isSupervisor` function defined as part of that logic

C-gyorfi (approved):

PR #411, opened 2023-10-11
Remove tests directory before cloning it

richardcane:
CircleCI runner seems to already have the tests directory sometimes - remove it and then clone it. Might be something to do with how the node orb is caching dependencies.

https://app.circleci.com/pipelines/github/ministryofjustice/bichard7-next-ui/3425/workflows/5fa3c89e-e8cf-4ce5-a46f-42359e27fdb0/jobs/22377


richardcane:
We're testing a better solution (hopefully) to this issue - just trying to confirm whether it's to do with cache restoration

PR #423, opened 2023-10-20
Don't lock resolved triggers or exception when viewing case details page

monotypical:
This PR:
* ensures that only unresolved triggers and exceptions are locked when viewing the case details page
* adds jest tests for the above
* adds cypress e2e tests for the above


richrace (commented):
The height of the "Resolved" icon is triggering for me 😅 

276981235-068f55aa-a6fd-40d8-aaf0-36e627416117

monotypical:
> The height of the "Resolved" icon is triggering for me 😅
> 
> 276981235-068f55aa-a6fd-40d8-aaf0-36e627416117

Fixed in https://github.com/ministryofjustice/bichard7-next-ui/pull/423/commits/2750aef4e11306abc28673b0a238df09e8403a11

ian-antking (approved):

PR #430, opened 2023-10-26
Remove previousPath query string parameter from the old Bichard URL

emadkaramad:
This PR is to remove the `previousPath` query string parameter from the the old Bichard URL in Switch to Old Bichard button.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,14 +188,14 @@ jobs:
     parallelism: 5
     steps:
       - setup
+      - run:
+          command: ./.circleci/scripts/install-cypress.sh 2>&1
+          background: true
       - run: npm run install:assets
       - fetch_images:
           IMAGES: nodejs-20-2023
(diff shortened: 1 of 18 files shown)


richrace:
We should make the same change here as well:
* https://github.com/ministryofjustice/bichard7-next-ui/blob/main/src/pages/feedback.tsx#L133 
* https://github.com/ministryofjustice/bichard7-next-ui/blob/main/src/components/PhaseBanner/PhaseBanner.tsx#L15

Keep it consistent behaviour.

emadkaramad:
> We should make the same change here as well:
> 
> * https://github.com/ministryofjustice/bichard7-next-ui/blob/main/src/pages/feedback.tsx#L133
> * https://github.com/ministryofjustice/bichard7-next-ui/blob/main/src/components/PhaseBanner/PhaseBanner.tsx#L15
> 
> Keep it consistent behaviour.

Good shout! Will update it

PR #445, opened 2023-11-07
Error prompt for offence code HO100306

BushraAbdullahi:
In this PR:

- Implemented the error prompt component when there is an HO100306 error in the offence code
- Wrote logic and tests to ensure error prompts only appear on the relevant offence
- Wrote logic and tests to ensure that no error prompts appear after exceptions have been marked as manually resolved

image


cypress/e2e/court-cases/court-case-details/ExceptionHandlerPrompt.cy.ts
```diff
@@ -34,19 +35,77 @@ describe("ExceptionHandlerPrompt", () => {
...
+      cy.get("ul.moj-sub-navigation__list").contains("Offences").click()
+      cy.get(".govuk-link").contains("Aid and abet theft").click()
+      cy.get(".qualifier-code-table").contains("XX")
+      cy.get(".qualifier-code-table .error-prompt-message").contains(ErrorMessages.QualifierCode)
+      cy.get(".offences-table").contains("YY10XYZXX")
+      cy.get(".offences-table .error-prompt-message").contains(ErrorMessages.HO100306_error_prompt)
+      cy.get("#exceptions-tab").contains("Exceptions").click()
+      cy.get("button").contains("Mark as manually resolved").click()
+      cy.get("button").contains("Resolve").click()
+      cy.get(".qualifier-code-table .error-prompt-message").should("not.exist")
```
monotypical, line 76: I think it's worth navigating back to the offence here to make sure we're looking at the right page before checking that things don't exist. If we changed marking as manually resolved in the future to take us to some other page, this test would still pass even though it wasn't checking what we intended

src/types/ErrorMessages.ts
```diff
@@ -1,5 +1,7 @@
 enum ErrorMessages {
-  QualifierCode = "This code could not be recognised, contact the courts to verify the correct information and report to the Bichard 7 team. Resolve exception by manually updating the PNC with correct code."
+  QualifierCode = "This code could not be recognised, contact the courts to verify the correct information and report to the Bichard 7 team. Resolve exception by manually updating the PNC with correct code.",
+  HO100251_error_prompt = "This code is not valid, contact the courts to correct offence code.",
```
monotypical, line 3: This name would be better camel cased to `HO100251ErrorPrompt` to be consistent with `QualifierCode`, and the same with `HO100306_error_prompt`

monotypical (commented):
I've left a couple of comments, other than that looks good!

monotypical (dismissed):

monotypical (approved):

PR #446, opened 2023-11-08
Bicaws 3134 update exception and trigger handlers filter behaviour

C-gyorfi:
### Context
During UR, we found that the combination of filters can result in odd behaviour, specifically combining resolution status, reason code and user permissions. 

### Changes
- Refactor list court cases function to ensure users can filter and see cases that their concern and have permission to see
- Extracted a `filterByReasonAndResolutionStatus` function to generate conditions based on user permission and applied reason/resolution status filter and cross-test these:
  - These filters are tightly coupled as applying one affects the others. For instance, a trigger handler filtering for resolved cases should see cases with resolved triggers only
- Added a test matrix for testing the combination of user permission, case state and reason filters
- Moved test from `listCourtCases` integration tests to test matrix where it makes sense

### Guidance for review
- I'd recommend looking through the test matrix to see if the filtering behaviour makes sense


test/services/listCourtCases/listCourtCases.integration.test.ts
C-gyorfi, line 1: I think we can also remove 
`describe("Filter cases having reason", () => {`
from this file

C-gyorfi: `Filter cases by case state` also covered/can be covered in the new test matrix

test/services/listCourtCases/listCourtCases.integration.test.ts
```diff
@@ -933,275 +922,6 @@ describe("listCourtCases", () => {
...
-      expect(isError(result)).toBeFalsy()
-      const { result: cases } = result as ListCourtCaseResult
-
-      expect(cases).toHaveLength(3)
-      expect(cases[0].errorId).toBe(2)
-      expect(cases[1].errorId).toBe(3)
-      expect(cases[2].errorId).toBe(4)
-    })
-
-    it("Should filter cases with all reasons", async () => {
```
C-gyorfi, line 1031: Can we also test this in the test matrix (multiple reason options selected)?

test/services/listCourtCases/filterByResolutionStatus.integration.test.ts
```diff
@@ -0,0 +1,727 @@
...
+        filters: {
+          caseState: "Resolved",
+          reasonCode: "TRP"
+        },
+        user: exceptionHandler,
+        expectedCases: ["Exceptions Resolved by exceptionHandler/Trigger Resolved by triggerHandler"]
+      },
+      {
+        description: "Should only see unresolved triggers when case state is not set as a trigger handler",
+        filters: {
```
C-gyorfi, line 645: In this case we could just set filters: {}

test/services/listCourtCases/filterByResolutionStatus.integration.test.ts
```diff
@@ -0,0 +1,727 @@
...
+        expectedCases: [
+          "Exceptions Resolved by exceptionHandler/Trigger Unresolved",
+          "Exceptions Unresolved/Trigger Unresolved",
+          "No exceptions/Bails Trigger Unresolved",
+          "Exceptions Unresolved/Bails Trigger Unresolved"
+        ]
+      },
+      {
+        description: "Should only see unresolved exceptions when case state is not set as an exceptions handler",
+        filters: {
```
C-gyorfi, line 659: Same here

emadkaramad (approved):

PR #462, opened 2023-11-16
Tweak the seed script

richrace:


test/helpers/createDummyCase.ts
```diff
@@ -32,20 +33,33 @@ export default async (
...
     messageId: uuidv4(),
     orgForPoliceFilter: orgCode,
     errorLockedByUsername: !isResolved && hasExceptions && randomBoolean() ? randomUsername() : null,
-    triggerLockedByUsername: !isResolved && hasTriggers && randomBoolean() ? randomUsername() : null,
+    triggerLockedByUsername:
+      !isResolved && hasUnresolvedTriggers && randomBoolean() && triggersExist ? randomUsername() : null,
     phase: 1,
     errorStatus: hasExceptions ? "Unresolved" : "Resolved",
-    triggerStatus: hasTriggers ? "Unresolved" : "Resolved",
+    triggerStatus: hasUnresolvedTriggers ? "Unresolved" : "Resolved",
```
C-gyorfi, line 62: We should set this to NULL when there are no triggers, same with errorStatus

test/helpers/createDummyCase.ts
```diff
@@ -58,8 +58,8 @@ export default async (
     triggerLockedByUsername:
       !isResolved && hasUnresolvedTriggers && randomBoolean() && triggersExist ? randomUsername() : null,
     phase: 1,
-    errorStatus: hasExceptions ? "Unresolved" : "Resolved",
-    triggerStatus: hasUnresolvedTriggers ? "Unresolved" : "Resolved",
+    errorStatus: exceptionCount === 0 ? null : hasExceptions ? "Unresolved" : "Resolved",
```
C-gyorfi, line 61: I don't think we ever seed resolved exceptions, question is, should we?

richrace: Good spot, I think we should.

C-gyorfi (approved):

PR #464, opened 2023-11-20
Fix sticky order by in case list

richrace:
Also,  refactor out the super long reducer for filtering cases.


src/features/CourtCaseFilters/reducers/filters.ts
```diff
@@ -0,0 +1,156 @@
+import type { Filter, FilterAction } from "types/CourtCaseFilter"
+import { caseStateLabels } from "utils/caseStateFilters"
+
+const handleAddingFilters = (newState: Filter, action: FilterAction) => {
```
monotypical, line 4: this is tidier and nicer than before

src/features/CourtCaseFilters/reducers/filters.ts
```diff
@@ -0,0 +1,156 @@
...
+    case "myCases": {
+      newState.myCasesFilter.value = action.value
+      newState.myCasesFilter.label = action.value ? "Cases locked to me" : undefined
+      newState.myCasesFilter.state = "Selected"
+      break
+    }
+    case "reason": {
+      // React might invoke our reducer more than once for a single event,
+      // so avoid duplicating reason filters
+      if (newState.reasonFilter.filter((reasonFilter) => reasonFilter.value === action.value).length < 1) {
```
monotypical, line 49: interesting that this is what was causing the problem, is it reproducible enough that we could add a cypress test for it?

monotypical (approved):

src/features/CourtCaseFilters/reducers/filters.ts
richrace: It was in the original reducer: https://github.com/ministryofjustice/bichard7-next-ui/pull/464/files#diff-88864ffabf9762a6197ac31ffe5d58b65eddafdda9162135bdbf303a9aea7208L64-L65

🤷‍♂️

src/features/CourtCaseFilters/reducers/filters.ts
richrace: Thank you

PR #498, opened 2023-12-08
Improve Case list row styling

BushraAbdullahi:
- Fixed the inconsistent borders. They were showing up between some rows only
- Changed the border colour and grey rows to adhere with the figma guidelines
- Refactored the CourtCaseListEntry component
- fixed the alignment of the note preview button
- implemented top vertical alignment of each row to adhere with figma guidelines
- Refactored unlocking.cy.ts to use helper functions


src/features/CourtCaseList/CourtCaseListEntry/CourtCaseListEntry.tsx
```diff
@@ -61,70 +52,49 @@ const CourtCaseListEntry: React.FC<Props> = ({
...
+        unlockPath={unlockCaseWithReasonPath("Trigger", `${errorId}`)}
+      />
+    )
   }
 
-  const classes = useCustomStyles()
-
   return (
-    <>
+    <tbody>
```
richrace, line 82: I don't like the repeated `<tbody>` within a single `<table>` and `<thead>`.

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tbody

src/features/CourtCaseList/CourtCaseListEntry/CaseDetailsRow/CaseDetailsRow.tsx
```diff
@@ -1,138 +1,76 @@
 import ConditionalRender from "components/ConditionalRender"
 import DateTime from "components/DateTime"
-import {
-  filterUserNotes,
-  getMostRecentNote
-} from "features/CourtCaseList/CourtCaseListEntry/CaseDetailsRow/CourtCaseListEntryHelperFunction"
+import { filterUserNotes } from "features/CourtCaseList/CourtCaseListEntry/CaseDetailsRow/CourtCaseListEntryHelperFunction"
```
richrace, line 3: I like this refactor! 🌟

styles/page.scss
```diff
@@ -3,4 +3,24 @@
...
+    tr:not(:last-child) td {
+      border-bottom: none;
+    }
+  }
+  tbody:nth-child(odd){
+    tr{
+      background-color: $b7-light-grey-color;
+    }
+  }
+  .note-preview-row {
```
richrace, line 21: Nitpick: Styling of the CSS

richrace (dismissed):

styles/page.scss
richardcane: Do we need a linter for this?

src/features/CourtCaseList/CourtCaseListEntry/CourtCaseListEntry.tsx
richrace: I see now why it needs to be a `<tbody>` 

Screenshot 2023-12-14 at 16 37 41

styles/page.scss
richrace: Always up for a linter 😅

emadkaramad (dismissed):

cypress/e2e/court-cases-index/pagination.cy.ts
```diff
@@ -22,6 +22,7 @@ describe("Pagination", () => {
     cy.get("h1")
 
     cy.checkA11y(undefined, a11yConfig, logAccessibilityViolations)
+    console.log(logAccessibilityViolations)
```
monotypical, line 25: does this need to be left in? the cypress GUI will show a list of accessibility violations already

cypress/e2e/court-cases-index/unlocking.cy.ts
```diff
@@ -94,30 +103,50 @@ describe("Case unlocked badge", () => {
     loginAndGoToUrl()
 
     // Exception lock
-    cy.get(`tbody tr:nth-child(1) .locked-by-tag`).get("button").contains("Bichard Test User 01").should("exist")
-    cy.get(`tbody tr:nth-child(1) img[alt="Lock icon"]`).should("exist")
+    cy.get("tbody")
```
monotypical, line 106: this quite complicated check that the lock tag exists gets repeated a few times, we could extract it to a helper function like is done for `unlockCase` to make changes to it easier

src/features/CourtCaseList/CourtCaseList.tsx
```diff
@@ -29,23 +28,23 @@ const CourtCaseList: React.FC<Props> = ({ courtCases, order = "asc" }: Props) =>
...
     .join("&")
 
-  const classes = useCustomStyles()
-
   return courtCases.length === 0 ? (
     <Paragraph>{"There are no court cases to show"}</Paragraph>
   ) : (
-    <Table head={<CourtCaseListTableHeader order={order} />}>
-      {courtCases.map((courtCase, index) => (
+    <table className="cases-list">
```
monotypical, line 34: We use `Table` from `govuk-react` in quite a few places, e.g. the rows are all `Table.Row`s. It would be nice to have consistency across them, and the govuk table handles layout nicely for us. Is it possible to use the govuk-react `Table` here or do we need to use the plain HTML table?

monotypical (commented):
A few suggestions, otherwise looks good!

cypress/e2e/court-cases-index/pagination.cy.ts
BushraAbdullahi: I forgot to remove it. thanks for spotting

src/features/CourtCaseList/CourtCaseList.tsx
BushraAbdullahi: we are using 'table' because we don't want to use the styling applied from 'Table'

monotypical (approved):
LGTM

PR #512, opened 2024-01-09
BICAWS7 653  - Exceptions  Confirmation Page

Senny10:
This request adds two versions of a confirmation page when an exception is submitted with or without editing. 

image

image


src/pages/court-cases/[courtCaseId]/submit.tsx
```diff
@@ -0,0 +1,181 @@
...
+  csrfToken: string
+  previousPath: string | null
+  hasAmendments?: boolean
+  amendments?: string
+}
+const SubmitCourtCasePage: NextPage<Props> = ({
+  courtCase,
+  user,
+  previousPath,
+  hasAmendments = false,
```
Comment (C-gyorfi) [inline src/pages/court-cases/[courtCaseId]/submit.tsx:125]: This prop could be removed, I'll explain in next comment ->

src/pages/court-cases/[courtCaseId]/submit.tsx
```diff
@@ -0,0 +1,181 @@
...
+          </BackLink>
+          <HeaderContainer id="header-container">
+            <HeaderRow>
+              <Heading as="h1" size="LARGE" aria-label="Submit Exception(s)">
+                {"Submit Exception(s)"}
+              </Heading>
+            </HeaderRow>
+          </HeaderContainer>
+
+          <ConditionalRender isRendered={hasAmendments}>
```
Comment (C-gyorfi) [inline src/pages/court-cases/[courtCaseId]/submit.tsx:155]: Could use 
```suggestion
          <ConditionalRender isRendered={amendments.length > 0}>
```
so `hasAmendments` is not needed

C-gyorfi (commented):
Looks good 🎉  I've added a few suggestions

cypress/e2e/court-cases/edit-fields-with-exceptions/nextHearingDate.cy.ts
```diff
@@ -3,6 +3,11 @@ import dummyAho from "../../../../test/test-data/error_list_aho.json"
 import hashedPassword from "../../../fixtures/hashedPassword"
 import { verifyUpdatedMessage } from "../../../support/helpers"
 
+function submitAndConfirmExceptions() {
+  cy.get("button").contains("Submit exception(s)").click()
+  cy.url().should("match", /\/bichard\/court-cases\/[0-9]+\/submit$/)
+  cy.get("button").contains("Submit exception(s)").click()
```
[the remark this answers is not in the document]
Senny10: The design shows: 
image
I need help with styling please.

src/pages/court-cases/[courtCaseId]/submit.tsx
Comment (Senny10) [reply src/pages/court-cases/[courtCaseId]/submit.tsx:155]: The reason `hasAmendments` is needed is because amendments is initialized `amendments: "{}"`. So `amendments.length` will always return 2. I would be happy to discuss alternative approaches.

emadkaramad (approved):

PR #524, opened 2024-01-17
Bicaws7 360/asn editable field

tausifpatel:
### [Exception Handler - ASN (Editable Field)](https://dsdmoj.atlassian.net/jira/software/c/projects/BICAWS7/boards/1368?selectedIssue=BICAWS7-360)

Make ASN field editable for the following scenarios
- The case must have at least one exception
- Current user has the permission to edit exceptions
- The case is locked to the current user
- Error status is unresolved.
- Case is in Phase 1 (Hearing outcome)


src/features/CourtCaseDetails/Tabs/Panels/DefendantDetails.tsx
```diff
@@ -13,38 +13,98 @@ import ErrorPromptMessage from "components/ErrorPromptMessage"
...
+  const [isValidAsn, setIsValidAsn] = useState<boolean>(true)
+  const handleAsnChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
+    const asn = e.target.value.toUpperCase()
+    setIsValidAsn(isAsnFormatValid(asn))
+    amendFn("asn")(asn)
+  }
+
+  const asnFormGroupError = isValidAsn ? "" : "govuk-form-group--error"
+  const isAsnEditable =
+    courtCase.canUserEditExceptions && courtCase.aho.Exceptions.length > 0 && courtCase.phase === Phase.HEARING_OUTCOME
```
C-gyorfi, line 61: `courtCase.aho.Exceptions.length > 0`

 don't we need to check if it is an ASN exception?

tausifpatel: No. As long as there is an unresolved exception, we want to make it editable.

src/services/dto/courtCaseDto.ts
```diff
@@ -13,6 +18,10 @@ export const courtCaseToDisplayPartialCourtCaseDto = (courtCase: CourtCase): Dis
     errorLockedByUsername: courtCase.errorLockedByUsername,
     errorReport: courtCase.errorReport,
     errorStatus: courtCase.errorStatus ?? null,
+    canUserEditExceptions:
```
C-gyorfi, line 21: I wonder if `canUserEditExceptions` could be defined elsewhere, it feels strange to pass current users as a dependency for court case DTO

src/features/CourtCaseDetails/Tabs/Panels/DefendantDetails.tsx
C-gyorfi: Then I think its redundant, `.canUserEditExceptions` already checks if error status is unresolved(should only be unresolved if there are unresolved exceptions) 
```
  const isAsnEditable =
    courtCase.canUserEditExceptions && courtCase.phase === Phase.HEARING_OUTCOME
```

src/services/dto/courtCaseDto.ts
richrace: You've got everything you need before you get to the UI (still on the server). So it's encapsulated the DTO and you don't need to expensive lookups.

richrace: So I think that's the right place for it be.

src/features/CourtCaseDetails/Tabs/Panels/DefendantDetails.tsx
tausifpatel: `courtCase.aho.Exceptions.length > 0`  is removed and updated an e2e test

C-gyorfi (dismissed):

richrace (approved):

PR #544, opened 2024-02-08
Bicaws7-2467: Add sylelint

ian-antking:


monotypical (approved):
hadn't come across stylint before, this looks great 🎉

PR #546, opened 2024-02-09
Remove 'Case list' header and padding from case list page

BushraAbdullahi:
image


src/pages/index.tsx
```diff
@@ -281,6 +281,7 @@ const Home: NextPage<Props> = (props) => {
         <CurrentUserContext.Provider value={currentUserContext}>
           <Layout bichardSwitch={{ display: true, displaySwitchingSurveyFeedback }}>
             <Main />
+            <h1 className="hidden-header">{"Case list"}</h1>
```
richrace, line 284: 1. Why do we need this line?
2. If we do need it, why is it off the page instead of:
```css
display: none;
visibility: hidden;
```

emadkaramad: There is an accessibility rule that there must be one H1 element on the page.

- with `display: none`, the element is removed from the layout, so Cypress won't be able to see the element.
- with `visibility: hidden`, the element is in the layout and maintains its position and size (which we don't want). (also not sure if Cypress can see the element in the case)

So we used negative fixed positioning to move the element outside the viewport. It's now visible to Cypress (and screen readers) but not visible to users.

src/features/CourtCaseFilters/CourtCaseFilterWrapper.tsx
```diff
@@ -18,44 +16,38 @@ const CourtCaseFilterWrapper: React.FC<Props> = ({
   paginationBottom
 }: Props) => {
   const [areAppliedFiltersShown, setAreAppliedFiltersShown] = useState(false)
-  const classes = useCustomStyles()
   return (
     <>
-      <div className={`${classes["top-padding"]} moj-filter-layout`}>
```
monotypical, line 24: looking at the MoJ filter pattern examples, they all have a containing div with the `moj-filter-layout` class, whereas this PR has removed the parent`moj-filter-layout` div. Do we need it?

BushraAbdullahi: My bad, I removed the whole div instead of removing just the class "top-padding"

PR #550, opened 2024-02-12
Fix submit exceptions button

tausifpatel:
### Disable or enable the button in following scenarios to prevent users from submitting cases with empty editable fields when exceptions occur.

[**User story** 
](https://dsdmoj.atlassian.net/jira/software/c/projects/BICAWS7/boards/1368?selectedIssue=BICAWS7-2505)

Make submit exception(s) button disabled:
- When all the editable fields are empty
- When one of the editable fields is empty

Make submit exception(s) button enabled:
- When all the editable fields are filled in.
- For ASN, the entered value should be valid.
- If an ASN exception is not raised but any other exception occurs, the ASN editable field can be left empty when submitting the case.


src/utils/editableFieldsValidationError.ts
```diff
@@ -0,0 +1,88 @@
...
+// return true when field is empty
+// return true when invalid value is entered
+
+///// If there is no exception
+// return false when field is empty
+// return true when invalid value is entered
+
+// if (amendments.asn) {
+//   return !isAsnFormatValid(amendments?.asn as string) || isEmptyOrWhiteSpace(amendments?.asn as string)
+// }
```
richrace, line 24: Is this meant to be here?

src/utils/editableFieldsValidationError.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+const editableFieldsValidationError = (courtCase: DisplayFullCourtCase, amendments: AmendmentRecords): boolean => {
+  const exceptions = courtCase.aho.Exceptions
+  if (exceptions.length === 0) {
+    return false
+  }
+
+  return (
+    // Object.entries(amendments).length === 0 ||
```
richrace, line 81: Another comment

src/utils/editableFieldsValidationError.ts
```diff
@@ -0,0 +1,88 @@
...
+// return true when invalid value is entered
+
+///// If there is no exception
+// return false when field is empty
+// return true when invalid value is entered
+
+// if (amendments.asn) {
+//   return !isAsnFormatValid(amendments?.asn as string) || isEmptyOrWhiteSpace(amendments?.asn as string)
+// }
+const asnValidationError = (exceptions: Exception[], amendments: AmendmentRecords): boolean => {
```
richrace, line 25: I would like if each error would in it's own file

richrace (changes requested):

src/utils/editableFieldsValidationError.ts
tausifpatel: Removed comments

src/utils/editableFieldsValidationError.ts
tausifpatel: Separated into each file

richrace (approved):
LGTM 👍

PR #561, opened 2024-02-25
Bicaws7 2538 case details whitespace reduction

Senny10:
- Removed page header (“Case Details”) and moved the defendant name to the top with H1 text size.
- Trigger and exception locks are displayed on one line, with ‘Exceptions’ and ‘Triggers’ in bold. 
- Removed the ‘Urgent’ badge from next to the defendant's name.

Before:
image
After:
image


src/components/Header/HeaderRow.ts
```diff
@@ -3,9 +3,13 @@ import styled from "styled-components"
...
-  align-items: flex-start;
+  align-items: center;
+  gap: 12px;
+  magin: 10px;
+  padding: 10px;
   & > #exceptions-locked-tag {
-    padding-top: 10px;
+  }
+  & > #trigger-locked-tag {
   }
```
richrace, line 13: I don't think we need this

richrace (approved):

PR #562, opened 2024-02-26
Pinned package version as breaks UI updates

joe-fol:
Quick win to fix failing npm update, will circle back in a week and test this has been updated upstream


config/.ncurc.js
richrace, line 1: Do forgot to delete this?

joe-fol: Yeah, alot happened yesterday not gonna lie, I'll remove now

bjpirt (approved):

PR #568, opened 2024-02-28
Only display searched for reasons on a filtered case list 

tausifpatel:
- Display only the reasons searched for on a filtered case list in the **Reason** column.
- Shorthand trigger and exception codes will not be searchable 

**Screenshots**
Court case list without applying filter:
Screenshot 2024-02-28 at 4 32 45 pm

Court case list after applying filter:
Screenshot 2024-02-28 at 4 32 35 pm

Court case list when filtered with shorthand code:
Screenshot 2024-02-28 at 4 33 16 pm


src/services/listCourtCases.ts
```diff
@@ -130,11 +130,15 @@ const listCourtCases = async (
...
-          reason: reasonCode
+        qb.where("trigger.trigger_code ilike :trigger", {
+          trigger: reasonCode
         })
+          .orWhere("courtCase.error_report ilike :firstException || '%'", {
+            firstException: reasonCode + "||"
+          })
+          .orWhere("courtCase.error_report ilike '%' || :exception || '%'", {
+            exception: " " + reasonCode + "||"
+          })
```
richrace, line 141: Could you use template literals, Eg
```typescript
const string = `${var} hello`
```

Trying to read this is hard.

test/services/listCourtCases/filterByReasonAndResolutionStatus.integration.test.ts
richrace, line 1: Is there a reason why the tests have been altered/deleted instead of fixed?

test/services/listCourtCases/listCourtCases.integration.test.ts
richrace, line 1: Same as above

src/services/listCourtCases.ts
tausifpatel: Changed into template literals.

test/services/listCourtCases/filterByReasonAndResolutionStatus.integration.test.ts
tausifpatel: When a user inputs a shorthand code instead of the full reason code, no court cases are returned. This results in an empty court case list array, hence the expectedCases array has 0 elements.

test/services/listCourtCases/listCourtCases.integration.test.ts
tausifpatel: On line 865, the inputted reason code is 2222. As we no longer accept shorthand reason codes, we get an empty case list. This means cases[0] and cases[1] no longer exist.

test/services/listCourtCases/filterByReasonAndResolutionStatus.integration.test.ts
richrace: We should change the description of this test and add a test for full reason code instead.

test/services/listCourtCases/listCourtCases.integration.test.ts
richrace: Same as above.

ian-antking (approved):

PR #597, opened 2024-04-08
Pinned date-fns-tz and eslint due to breaking updates

joe-fol:


.ncurc.js
```diff
@@ -12,7 +12,7 @@
   - next
     - 13.4.13 causes failures with fetch
 */
-const pinned = ["styled-components", "date-fns"]
+const pinned = ["styled-components", "date-fns", "date-fns-tz", "eslint"]
```
richrace, line 15: `date-fns-tz` supports to v3 https://github.com/marnusw/date-fns-tz/releases/tag/v3.0.0 ! 

We should do a PR to remove `"date-fns", "date-fns-tz"` as it requires breaking changes 🙂

richrace (approved):

PR #600, opened 2024-04-11
Set up and use users more sensibly and don't wipe the database

bjpirt:
This is a bit of an epic refactor of the Cypress and integration tests which:
 - stops them from wiping out the users in the database (and making your local Bichard unusable)
 - standardises setting up new users so you don't have to manually create them in all of the tests
 - creates a shortcut which creates (if necessary) the user, logs in as them and visits a URL

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,7 +224,7 @@ jobs:
       image: ubuntu-2204:current
       docker_layer_caching: true
     resource_class: xlarge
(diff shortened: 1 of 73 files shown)


richardcane:
Just looked this over with Ben, one other idea would be to set up `loginAndVisit` as a custom cypress command.

richardcane (approved):

PR #626, opened 2024-05-01
Feature flag editable fields, submit exceptions button and exceptions icons

tausifpatel:
User story - https://dsdmoj.atlassian.net/jira/software/c/projects/BICAWS7/boards/1368?selectedIssue=BICAWS7-2750

### DIFF
--- a/src/utils/getTabDetails.test.ts
+++ b/src/utils/getTabDetails.test.ts
@@ -295,6 +295,33 @@ describe("getTabDetails", () => {
     }
   )
 
+  it.each([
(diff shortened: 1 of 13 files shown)


src/utils/getTabDetails.test.ts
richrace, line 1: I would to see a test here that sets  `const exceptionsEnabled = false`

tausifpatel: Added a test

richrace (approved):

PR #633, opened 2024-05-16
Changed ip used on port 3000

joe-fol:
the IP was changed yesterday that next was run on and the goss test didn't run till it hit code build. This PR just changes what we expect in goss


chubberlisk (approved):
:shipit:

PR #634, opened 2024-05-23
Enable ASN input field to automatically format input into sections within the text field

tausifpatel:
User story - https://dsdmoj.atlassian.net/jira/software/c/projects/BICAWS7/boards/1368?selectedIssue=BICAWS7-1465

Features added:
- Divide ASN into sections when the user enters it into the ASN field, for example: 11/01ZD/01/00000448754K.
- Divide ASN into sections when user pastes ASN in the ASN field
- Remove divisions (slashes) when user copies ASN from the ASN field
- Allow users to provide shorter version of ASN
- Display a checkmark icon next to the input box when the typed ASN is in a valid format.

**Screenshots:**
**Longer version:**
Screenshot 2024-05-23 at 10 56 29 AM
**Shorter version:**
Screenshot 2024-05-23 at 10 57 14 AM
**Error:**
Screenshot 2024-05-23 at 10 56 39 AM

### DIFF
--- a/src/services/Asn.ts
+++ b/src/services/Asn.ts
@@ -1,6 +1,8 @@
 const intOrString = (input: string): string => (input.match(/^\d*$/) ? input : "")
 
 class Asn {
+  public asn: string
+
   public year: string | undefined
 
   public force: string | undefined
@@ -12,6 +14,7 @@ class Asn {
   public sequence: number | undefined
 
   constructor(asn: string) {
+    this.asn = asn
     const asnString = asn.replace(/\//g, "")
     if (asnString) {
       this.year = asnString.slice(0, 2)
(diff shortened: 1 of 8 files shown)


src/services/Asn.ts
```diff
@@ -37,6 +41,51 @@ class Asn {
       this.sequence ?? "".toString().padStart(11, "0")
     }${this.checkCharacter()}`
   }
+
+  static yearChars = 2
+
+  static forceAndUnitChars = 4
+
+  static systemChars = 2
```
richrace, line 49: Suggestion: I would put the magic numbers after the `class Asn` and before `public asn: string`. I think it's more typical for languages.

richrace (approved):

src/services/Asn.ts
tausifpatel: Rearranged variables

richrace (approved):

PR #649, opened 2024-06-20
WIP: Bicaws7 2883 display latest notes when autosaving editable fields

C-gyorfi:


richrace:
#648 includes the Fix sticky error on the next hearing location
#650 includes the ability to update the notes with auto save

PR #654, opened 2024-06-24
Refactor AutoSave and add tests (and fix misc bugs)

richrace:
* Extract the function excludeSavedAmendments (with unit tests)
* Fix sticky Exceptions/Trigger filter chip (flakey cypress e2e test)
* Add cypress component test for AutoSave component
* Add cypress e2e tests for updating notes - proves the court case got updated
* Fix submit button ignores amendments already saved and add e2e cypress tests


cypress/e2e/case-details/handle-triggers-and-exceptions/edit-fields-with-exceptions/edit-asn.cy.ts
```diff
@@ -286,4 +329,47 @@ describe("ASN", () => {
   it("Should not display an editable field for ASN when exceptionsEnabled is false for user", () => {
     loginAndVisit("NoExceptionsFeatureFlag", "/bichard/court-cases/0")
   })
+
+  describe("when I submit resolved exceptions I should not the same value in the notes", () => {
```
C-gyorfi, line 333:
```suggestion
  describe("when I submit resolved exceptions I should not see the same value in the notes", () => {
```

cypress/e2e/case-details/handle-triggers-and-exceptions/edit-fields-with-exceptions/edit-next-hearing-location.cy.ts
```diff
@@ -404,4 +444,50 @@ describe("NextHearingLocation", () => {
     cy.get("#next-hearing-location-row .warning-icon").should("exist")
     cy.contains("Select valid Next hearing location").should("exist")
   })
+
+  describe("when I submit resolved exceptions I should not the same value in the notes", () => {
```
C-gyorfi, line 448:
```suggestion
  describe("when I submit resolved exceptions I should not see the same value in the notes", () => {
```

src/features/CourtCaseDetails/Sidebar/Exceptions.tsx
```diff
@@ -46,6 +48,12 @@ const Exceptions = ({ onNavigate, canResolveAndSubmit, stopLeavingFn }: Props) =
     stopLeavingFn(false)
   }
 
+  const amendmentsToUpdate = excludeSavedAmendments(
```
C-gyorfi, line 51: 🎉

src/utils/autoSave/excludeSavedAmendments.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+  })
+
+  if (!isEmpty(newValues)) {
+    map.set(amendmentField, newValues)
+  }
+}
+
+const excludeSavedAmendments = (
+  amendmentFields: AmendmentKeys[],
```
C-gyorfi, line 35: I wonder if it's necessary to pass the amendmentFields as an argument

src/components/EditableFields/AutoSave.tsx
```diff
@@ -104,3 +96,5 @@ export const AutoSave = ({
     </>
   )
 }
+
+export default AutoSave
```
C-gyorfi, line 100: Is export default the preferred way of exporting a function/component? I liked that export const would ensure we use the same name at import.

C-gyorfi (dismissed):
Awesome work, added some comments but nothing that would block the merge 👏

src/components/EditableFields/AutoSave.tsx
richrace: I prefer exporting the component. Maybe an ADR conversion or best practice? Decide as a team how we manage exporting React Components. There's nothing to stop using the export const function and having the default as well.

src/utils/autoSave/excludeSavedAmendments.ts
richrace: From the AutoSave point of view, it only wants to save a single value, e.g. `asn`

And from the `Submit Exceptions`, it removes the `offenceReasonSequence`, which gets added to the notes.

C-gyorfi (approved):

PR #662, opened 2024-07-01
BICAWS7-2916: Add Date Matching to OffenceMatcher

Harriet-Hall:
Factors the offence match logic in `OffenceMatcher` to  `isCaseMatch` and imports `generateCandidate` from core phase 1.

Adds dependency injection to `OffenceMatcher` so render logic is tested independently.

Changes `CoutCaseContext` so that dates are parsed as `Date` objects. We may want to do further refactor to remove ad-hoc `new Date()` parsing from the rest of the UI.


src/components/OffenceMatcher.tsx
```diff
@@ -1,26 +1,37 @@
...
 import offenceMatcherSelectValue from "utils/offenceMatcher/offenceMatcherSelectValue"
 import Badge, { BadgeColours } from "./Badge"
+import _isOffencePossibleMatch from "utils/offenceMatcher/isOffencePossibleMatch"
+import { PncOffence } from "@moj-bichard7-developers/bichard7-next-core/core/types/PncQueryResult"
 
 interface Props {
   offenceIndex: number
   offence: Offence
   isCaseLockedToCurrentUser: boolean
+  isOffencePossibleMatch?: (
```
richardcane, line 14: As discussed, we should refactor out the business logic from this component as a further next step.

richardcane (approved):

PR #679, opened 2024-07-23
Reason codes not displaying on case reason column

richrace:
* Add check for `reasonCodes` to narrow the search
* Only filter on triggers/exceptions when the given codes are only triggers/exceptions
* Stops 'andOr' search for triggers and exceptions
* Added tests for `reasonCodes`


test/services/listCourtCases/filterByReasonAndResolutionStatus.integration.test.ts
```diff
@@ -341,6 +341,152 @@ describe("Filter cases by resolution status", () => {
...
+        },
+        user: triggerHandler,
+        expectedCases: [
+          "Exceptions Unresolved/Bails Trigger Unresolved", // Sees this case as it has the reason code
+          "No exceptions/Bails Trigger Unresolved"
+        ]
+      },
+      {
+        description:
+          "Should see cases with unresolved exception when user is a exception handler and searches for TRPR0010",
```
tausifpatel, line 358: Should the title have HO100300 instead of trigger code?

richrace: 🤦‍♂️

tausifpatel (approved):

PR #688, opened 2024-07-30
Remove unneeded CSRF cookies

richrace:
* ~~Removes a lot of CSRF cookies~~
* ~~Doesn't remove the filtering CSRF cookies~~
* Removed all CSRF cookies
* Checked the forms for CSRF values
* Calls to the API doesn't need CSRF tokens as it checks the auth token


richrace:
Works in dev but not docker

ian-antking (approved):

PR #693, opened 2024-08-02
[SPIKE] Add trigger counts next the trigger groups

richrace:
Screenshot 2024-08-02 at 14 57 56

Expands text on hover or focused (while using tab)
Screenshot 2024-08-02 at 14 58 09


richrace:
This was a spike to see if it was possible. Further research is needed because we need to be clear about the number presents.

PR #694, opened 2024-08-02
Remove exceptions count icons when exceptions resolved manually

tausifpatel:
Context:

When exceptions are resolved manually, the exception count icons on tabs and the exception warning icons on offenses are still displayed. This confuses users, as these icons suggest that an action needs to be taken, even though they have already resolved the case manually.

Solution: 

Do not display any of the exceptions icons when exceptions resolved manually


richrace:
Is there an e2e test for this? Just want to check when we have submitted an exception, it doesn't say "Manually Resolved".

richrace (approved):

PR #699, opened 2024-08-09
Make the PNC sequence number field actually update the field

richrace:
This is a spike for if we need to update the PNC sequence number.


richrace:
This is not needed.

PR #701, opened 2024-08-12
Pin eslint libararies as peer deps using older version

joe-fol:


.ncurc.js
```diff
@@ -12,7 +12,14 @@
   - next
     - 13.4.13 causes failures with fetch
 */
-const pinned = ["styled-components", "date-fns", "date-fns-tz", "eslint"]
+const pinned = [
+  "styled-components",
+  "date-fns",
+  "date-fns-tz",
```
richrace, line 18: We don't need this now!

* https://github.com/marnusw/date-fns-tz/releases/tag/v3.0.0

emadkaramad (approved):

PR #702, opened 2024-08-12
update regex to include \r\n characters

Harriet-Hall:


src/services/addNote.ts
```diff
@@ -5,7 +5,7 @@ import CourtCase from "./entities/CourtCase"
 import insertNotes from "./insertNotes"
 
 const MaxNoteLength = 2000
-const notesRegex = new RegExp(`(.{1,${MaxNoteLength}})`, "g")
+const notesRegex = new RegExp(`(.|\\r|\\n){1,${MaxNoteLength}}`, "g")
```
emadkaramad, line 8: You can use `\s` instead which includes `.`, `\r`, and `\n`.

```suggestion
const notesRegex = new RegExp(`(\\s){1,${MaxNoteLength}}`, "g")
```

src/services/addNote.ts
```diff
@@ -32,7 +32,6 @@ const addNote = async (
       ValidationException: "Case is not locked by the user"
     }
   }
-
```
emadkaramad, line 35: We need an empty line after if blocks. This is for better readability.

richrace (changes requested):
Can you add tests for when the note has new lines to the existing test please?

`test/services/addNote.integration.test.ts`

emadkaramad (approved):

richrace (approved):

PR #719, opened 2024-09-03
Add resolvedByUsername to case list query

BushraAbdullahi:


src/services/listCourtCases.ts
```diff
@@ -134,26 +143,55 @@ const listCourtCases = async (
...
+        })
     }
   }
 
+  if (resolvedByUsername) {
+    let splitResolvedByUsername = resolvedByUsername.replace(/\*|\s+/g, "%")
+
+    if (!splitResolvedByUsername.endsWith("%")) {
+      splitResolvedByUsername = `${splitResolvedByUsername}%`
+    }
```
richrace, line 173: I think this should be refactored into a function as it's used twice, so we can test it independently.

What do you think?

test/services/listCourtCases/listCourtCases.integration.test.ts
```diff
@@ -539,6 +539,32 @@ describe("listCourtCases", () => {
...
+      expect(cases[2].errorResolvedBy).toStrictEqual("User Name03")
+
+      result = await listCourtCases(dataSource, { maxPageItems: 100, resolvedByUsername: "User Name01" }, testUser)
+      expect(isError(result)).toBe(false)
+      cases = (result as ListCourtCaseResult).result
+
+      expect(cases).toHaveLength(1)
+      expect(cases[0].errorResolvedBy).toStrictEqual("User Name01")
+    })
+  })
```
richrace, line 567: This should have its own test file, e.g. `test/services/listCourtCases/filterByDefendantName.integration.test.ts`

src/services/listCourtCases.ts
```diff
@@ -12,6 +11,16 @@ import filterByReasonAndResolutionStatus from "./filters/filterByReasonAndResolu
...
+function formatName(name: string) {
+  let splitName = name.replace(/\*|\s+/g, "%")
+
+  if (!splitName.endsWith("%")) {
+    splitName = `${splitName}%`
+  }
+
+  return splitName
+}
+
```
richrace, line 23: I think this should be refactored out into a function file so we can test it independently

test/services/listCourtCases/filterByResolvedByUsername.integration.test.ts
```diff
@@ -0,0 +1,78 @@
...
+  })
+
+  afterAll(async () => {
+    if (dataSource) {
+      await dataSource.destroy()
+    }
+  })
+
+  describe("search by resolvedByUsername", () => {
+    it("Should list cases that have been resolved by a specific user", async () => {
```
richrace, line 54: I would expect more test examples. For example, what happens if you filter using the wildcards?

test/services/listCourtCases/filterByResolvedByUsername.integration.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      let result = await listCourtCases(dataSource, { maxPageItems: 100, resolvedByUsername: "User" }, testUser)
+      expect(isError(result)).toBe(false)
+      let { result: cases } = result as ListCourtCaseResult
+
+      expect(cases).toHaveLength(3)
+      expect(cases[0].errorResolvedBy).toStrictEqual("User Name01")
+      expect(cases[1].triggerResolvedBy).toStrictEqual("User Name02")
+      expect(cases[2].errorResolvedBy).toStrictEqual("User Name03")
+
+      result = await listCourtCases(dataSource, { maxPageItems: 100, resolvedByUsername: "User Name01" }, testUser)
```
richrace, line 70: This should be a different test case.

src/helpers/formatName.ts
```diff
@@ -0,0 +1,9 @@
+export function formatName(name: string) {
```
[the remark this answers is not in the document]
BushraAbdullahi: apologies. something told me I needed to do it in another pr

test/services/listCourtCases/filterByResolvedByUsername.integration.test.ts
BushraAbdullahi: I was following the format of the other integration tests for listCourtCases

richrace: We should look to improve our code going forward. This specific test gets lost, so it becomes hard to read. We want our tests as simple as possible.

richrace (approved):
Nice! LGTM 👍

PR #724, opened 2024-09-09
Bicaws7-3076: listCases query select refactor

ian-antking:
Factors out the columns we select in the `listCourtCases` function. This should let us re-use the same function when we want to start exporting complete cases for reports.

Removes code written when we were going for a more literal reimplementation of the reports in legacy Bichard.

Moves some tests out of the `listCourtCases` integration tests, to help keep the size of that file down.


test/services/listCourtCases/filterCasesByResolvedDate.integration.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    const { result, totalCases } = (await listCourtCases(
+      dataSource,
+      { resolvedDateRange, caseState: "Resolved" },
+      testUser
+    )) as ListCourtCaseResult
+
+    console.log(
+      "Result: ",
+      result.map((r) => r.resolutionTimestamp)
+    )
```
richrace, line 78: Do we want `console.log` here?

test/services/listCourtCases/filterCasesByResolvedDate.integration.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  })
+
+  afterAll(async () => {
+    if (dataSource) {
+      await dataSource.destroy()
+    }
+  })
+
+  it("Returns resolved cases within given date range", async () => {
```
richrace, line 55: I would expect more tests.

richrace (approved):

